PR #15, opened 2021-04-09
Image delete and Jobmetadata validate routes

Geoffsc:
We need to add the functions from the ruby in JobMetadataValidator to JobMetadata.Validate but this is a start for now to discuss


api/src/routes/api.ts
```diff
@@ -65,8 +66,20 @@ router.get("/:category/:job/:image/:size", async function(req, res, next) {
...
+router.delete("/:category/:job/:image/*"), async function(req, res, next) {
+    //return ImageFile.delete();
+    let image: string = req.params.image;
+    let job = getJobFromRequest(req);
+    let imageObj = job.getImage(image);
+    if (imageObj !== null) {
+        imageObj.delete();
+        res.send(JSON.stringify( { status: 'ok' } ));
+    } else {
+        res.send(JSON.stringify( { status: 'image missing' } ));
```
demiankatz: This should also return a 404 status; not sure how to do that, but it's probably not hard.

api/src/routes/api.ts
```diff
@@ -65,8 +66,20 @@ router.get("/:category/:job/:image/:size", async function(req, res, next) {
     let size: string = req.params.size;
     let job = getJobFromRequest(req);
     let deriv = await job.getImage(image).derivative(legalSizes[size] ?? "THUMBNAIL");
-    console.log(deriv);
     res.sendFile(deriv);
 });
 
+router.delete("/:category/:job/:image/*"), async function(req, res, next) {
+    //return ImageFile.delete();
```
demiankatz: Unnecessary comment, I think.

api/src/routes/api.ts
```diff
@@ -65,8 +66,20 @@ router.get("/:category/:job/:image/:size", async function(req, res, next) {
...
+    let image: string = req.params.image;
```
demiankatz, line 76: We should sanitize/validate this parameter (along with all the others!). Not sure if there's a way to apply constraints to route parameters through the route definition, or if we need to add logic to do the work... but one way or another, that's a detail we should handle soon, since it's a security thing.

api/src/routes/api.ts
```diff
@@ -49,8 +49,9 @@ router.put("/:category/:job/ingest", function(req, res, next) {
 });
 
 router.put("/:category/:job", function(req, res, next) {
-    //res.send(JSON.stringify(getJobFromRequest(req).metadata.upda));
-    //res.send(JSON.stringify( { status: 'ok' } ));
+    let job = getJobFromRequest(req);
+    job.metadata.validate(job, req.params);
```
demiankatz: This doesn't seem to be complete yet... the idea is that the route should use the validate() method to figure out whether the incoming data is valid or not. If it's invalid, it needs to return a 400 status with an error message, and otherwise, it needs to save the valid data to disk.

api/src/models/ImageFile.ts
```diff
@@ -79,8 +79,18 @@ class ImageFile {
 
     }
 
-    delete() {
-
+    public delete() {
+        let fs = require("fs");
+        for (let size in this.sizeArray) {
```
demiankatz: This is not yet complete -- this appears to be only deleting JPEG derivatives, but it also needs to remove other types of derivatives (OCR, etc.) plus the original TIFF itself.

api/src/models/ImageFile.ts
```diff
@@ -7,6 +8,7 @@ class ImageFile {
         "MEDIUM": 640,
         "THUMBNAIL": 120
     };
+    sizeArray: Array<string> = ["LARGE","MEDIUM","THUMBNAIL"];
```
demiankatz: Rather than duplicating information here, can you just pull the keys from sizes instead of using a separate array?

api/src/models/JobMetadata.ts
```diff
@@ -43,6 +43,10 @@ class JobMetadata {
         return job.dir + '/ingest.lock';
     }
 
+    validate(job, params) {
+
```
demiankatz: We can talk about how validation works next week; I've refreshed my memory, and it is indeed a bit complicated. It might make sense to use a separate validator class rather than building it in... or it might not. We can discuss next week!

demiankatz (commented):
Thanks, @Geoffsc! I've left an initial round of comments, but of course I don't expect you to worry about any of them until next week. :-)

api/src/routes/api.ts
Geoffsc: Removed!

api/src/routes/api.ts
Geoffsc: Added 404

api/src/routes/api.ts
```diff
@@ -70,15 +70,16 @@ router.get("/:category/:job/:image/:size", async function(req, res, next) {
...
+    //Sanitize incoming parameters
     let image: string = req.params.image;
     let job = getJobFromRequest(req);
     let imageObj = job.getImage(image);
     if (imageObj !== null) {
         imageObj.delete();
         res.send(JSON.stringify( { status: 'ok' } ));
     } else {
-        res.send(JSON.stringify( { status: 'image missing' } ));
+        res.status(404).send("image missing");
```
demiankatz: The JSON response seems to have gotten broken when you fixed the 404 status; it's just returning the string "image missing" now, instead of the JSON object containing a status field.

demiankatz:
One other general comment: we have a mix of "TODO" and "TO DO" comments; we should standardize these so they're easier to search for all at once. I'd recommend "TODO" because it's more distinctive for search purposes.

api/src/models/ImageFile.ts
```diff
@@ -79,8 +78,22 @@ class ImageFile {
 
     }
 
-    delete() {
-
+    public delete() {
+        let fs = require("fs");
+        let files: Array<string> = [];
+        for (let size in Object.keys(this.sizes)) {
+            files.push(this.derivativePath(size, "jpg"));
```
demiankatz, line 141: I don't think you need the png/tiff/txt lines here -- we only create jpg derivatives in the sizes. This method just needs to undo the work done by other methods. You do still need to add the logic to remove OCR files and the original TIFF, however.

api/src/models/ImageFile.ts
Geoffsc: Updated!

api/src/models/ImageFile.ts
Geoffsc: Updated!

api/src/models/ImageFile.ts
Geoffsc: Updated!

api/src/models/ImageFile.ts
```diff
@@ -56,7 +56,7 @@ class ImageFile {
         var path = require('path');
         var dir = path.dirname(this.filename);
         var filename = this.basename(this.filename);
-        return dir + "/" + filename + "." + size + "." + extension.toLowerCase();
+        return dir + "/" + filename + "/" + size + "/" + extension.toLowerCase();
```
demiankatz: For consistency with the Ruby, I think you want:

`return dir + "/" + filename + "/" + size + "/" + filename + "." + extension.toLowerCase();`

(It uses the filename both as a directory for grouping things together, and as the actual filename for the generated derivative... I don't know that all this is strictly necessary, but if we want to keep it as it is, that other filename should be added back in).

api/src/models/ImageFile.ts
```diff
@@ -71,21 +71,50 @@ class ImageFile {
     }
 
     ocrDerivative() {
-
+        let fs = require("fs");
+        let png = this.derivativePath('ocr/pngs', 'png');
+        let { exec } = require("child_process");
+        if (fs.existsSync(png)) {
```
demiankatz: I think this should be `if (!fs.existsSync(png)) {` -- we want to create the file if it does not already exist.

api/src/models/ImageFile.ts
```diff
@@ -71,21 +71,50 @@ class ImageFile {
...
+            let tc_cmd = "#{config['textcleaner_path']} #{config['textcleaner_switches']} #{derivative('LARGE')} #{png}"
```
demiankatz: It looks like you copied the Ruby-style string embedding directly from the Ruby; this won't work as-is in Javascript. You'll need to pull these settings from the configuration object, and put the string together through concatenation. Also note that `derivative('LARGE')` is a call to get the path to the large derivative file.

api/src/models/ImageFile.ts
```diff
@@ -71,21 +71,50 @@ class ImageFile {
...
+                    console.log(`error: ${error.message}`);
+                    return;
+                }
+                if (stderr) {
+                    console.log(`stderr: ${stderr}`);
+                    return;
+                }
+                console.log(`stdout: ${stdout}`)
+            });
+        }
```
demiankatz, line 113: At this point, the Ruby code checks to see if the output file was successfully generated and the command line process did not return an error. We should probably do similar validation here.

api/src/models/ImageFile.ts
```diff
@@ -71,21 +71,50 @@ class ImageFile {
...
     }
 
     ocrProperties() {
-
+        let fs = require("fs");
+        var path = require('path');
+        let file = this.derivativePath('ocr/pngs', 'png');
+        let dir = path.dirname(file);
+        let content = 'tessedit_char_whitelist ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789.,;:!\'"()&$%-+=[]?<>' + "\xE2\x80\x9C\xE2\x80\x9D\xE2\x80\x98\xE2\x80\x99";
+        if (fs.existsSync(file)) {
```
demiankatz: Another one that should be inverted to `!fs.existsSync(file)`...

api/src/models/ImageFile.ts
```diff
@@ -71,21 +71,50 @@ class ImageFile {
...
+                console.log(`stdout: ${stdout}`)
+            });
+        }
     }
 
     ocrProperties() {
-
+        let fs = require("fs");
+        var path = require('path');
+        let file = this.derivativePath('ocr/pngs', 'png');
```
demiankatz: I think this is copied and pasted from the ocrDerivative() method, but for this you just want to create 'ocr/tesseract.config' inside the current job directory.

api/src/models/ImageFile.ts
```diff
@@ -71,21 +71,50 @@ class ImageFile {
...
+        let content = 'tessedit_char_whitelist ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789.,;:!\'"()&$%-+=[]?<>' + "\xE2\x80\x9C\xE2\x80\x9D\xE2\x80\x98\xE2\x80\x99";
+        if (fs.existsSync(file)) {
+            fs.writeFile(dir + '/ocr/tesseract.config', content, err => {
+                if (err) {
+                  console.error(err)
+                  return
+                }
+                //file written successfully
+              })
+        }
```
demiankatz, line 130: This method needs to return the name of the properties file that it created.

api/src/models/ImageFile.ts
demiankatz: Thanks. Nearly there, but for completeness, I think it's also worth deleting `this.derivativePath('ocr/pngs', 'png');` and `this.derivativePath('OCR-DIRTY', 'txt');` for consistency with the Ruby (though as we discussed, I think in practice that's never going to be needed).

demiankatz (changes requested):
Thanks for the latest updates; I've added a few new comments.

api/src/models/ImageFile.ts
```diff
@@ -74,29 +80,33 @@ class ImageFile {
         let fs = require("fs");
         let png = this.derivativePath('ocr/pngs', 'png');
         let { exec } = require("child_process");
-        if (fs.existsSync(png)) {
-            let tc_cmd = "#{config['textcleaner_path']} #{config['textcleaner_switches']} #{derivative('LARGE')} #{png}"
+        if (!fs.existsSync(png)) {
+            let tc_cmd = this.config().textcleanerPath() + this.config().textcleanerSwitches() + this.derivative('LARGE') + png;
```
demiankatz: You're going to need to add `+ " "` in between all of these things so the parameters don't run together.

api/src/models/ImageFile.ts
demiankatz: What you added above is a good start, but I think we want to do more than just log a message. We need to fail the build so the problem can be investigated... otherwise jobs may get ingested with no OCR, and we could easily fail to notice. I think the Javascript equivalent of Ruby's `raise` is `throw`.

api/src/models/ImageFile.ts
demiankatz: I also think you might need to `await` the call to `this.derivative()` because it's an `async` call...

demiankatz (changes requested):
Thanks! We're making progress -- just a couple more things from today's review.

api/src/models/ImageFile.ts
```diff
@@ -65,34 +65,52 @@ class ImageFile {
...
-        if (fs.existsSync(txt)){
+        let { exec } = require("child_process");
+        let deriv = await this.ocrDerivative();
+        if (!fs.existsSync(txt)){
             var path = this.basename(txt);
+            let ts_cmd = this.config().tesseractPath() + " " + deriv + " " + txt.slice(0, -4) + " " + this.ocrProperties();
+            exec(ts_cmd, (error, stdout, stderr) => {
+                if (error) {
+                    throw `error: ${error.message}`;
+                    return;
```
demiankatz: I think you might be able to refine this a little bit...

1.) I don't think it's necessary to `return` after a `throw`, because the `throw` will already cause you to exit the function.
2.) I don't know if you want to throw an error just because there is content in the stderr stream; it's possible that Tesseract might throw data to stderr even when successful (or not... I can't remember). The error parameter is the main thing to be worried about, since if it is non-null, there is definitely a problem. Maybe the approach should be to include the stderr output in the exception thrown when error is non-null, and to console.log it for informational purposes under other circumstances.

api/src/models/ImageFile.ts
```diff
@@ -65,34 +65,52 @@ class ImageFile {
...
         let png = this.derivativePath('ocr/pngs', 'png');
         let { exec } = require("child_process");
+        let deriv = await this.derivative('LARGE');
         if (!fs.existsSync(png)) {
-            let tc_cmd = this.config().textcleanerPath() + this.config().textcleanerSwitches() + this.derivative('LARGE') + png;
+            let tc_cmd = this.config().textcleanerPath() + " " + this.config().textcleanerSwitches() + " " + deriv + " " + png;
             exec(tc_cmd, (error, stdout, stderr) => {
                 if (error) {
-                    console.log(`error: ${error.message}`);
+                    throw `error: ${error.message}`;
```
demiankatz: Same comments from above apply here as well.

api/src/models/ImageFile.ts
```diff
@@ -56,29 +62,98 @@ class ImageFile {
...
 
     ocrProperties() {
-
+        let fs = require("fs");
+        var path = require('path');
+        let file = path.dirname(this.filename) + '/ocr/tesseract.config';
+        let dir = path.dirname(file);
+        let content = 'tessedit_char_whitelist ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789.,;:!\'"()&$%-+=[]?<>' + "\xE2\x80\x9C\xE2\x80\x9D\xE2\x80\x98\xE2\x80\x99";
+        if (!fs.existsSync(file)) {
+            fs.writeFile(dir + '/ocr/tesseract.config', content, err => {
```
demiankatz: I think you can just use `file` here instead of `dir + '/ocr/tesseract.config'`  -- right now you seem to be taking the string apart and then putting it back together again, but it should be possible to do less work. :-)

api/src/models/ImageFile.ts
```diff
@@ -56,29 +62,98 @@ class ImageFile {
...
+                if (err) {
+                  console.error(err)
```
demiankatz: This should probably be a `throw`.

demiankatz (commented):
Another round of comments. We're getting close! :-)

api/src/models/ImageFile.ts
```diff
@@ -76,15 +76,13 @@ class ImageFile {
...
+                    throw `error: ${error.message + " " + stderr}`;
                 } 
                 if (stderr) {
-                    throw `stderr: ${stderr}`;
-                    return;
+                    console.log(`stderr: ${stderr}`);
                 }
                 if (!fs.existsSync(txt)) {
-                    throw "Problem running textcleaner";
+                    console.log("Problem running textcleaner");
```
demiankatz: This should still be a throw -- if the file we expected to create does not exist after all is said and done, something is very wrong and we should consider that an exception.

api/src/models/ImageFile.ts
```diff
@@ -102,15 +100,13 @@ class ImageFile {
...
+                    throw `error: ${error.message + " " + stderr}`;
                 } 
                 if (stderr) {
-                    throw `stderr: ${stderr}`;
-                    return;
+                    console.log(`stderr: ${stderr}`);
                 }
                 if (!fs.existsSync(png)) {
-                    throw "Problem running textcleaner";
+                    console.log("Problem running textcleaner");
```
demiankatz: Ditto here: this should still be a throw.

demiankatz (changes requested):
I think there's been a slight overcorrection -- see comments below. Once that's addressed, I'm fine with having this merged. We obviously need to test it, but until we're in a position to do that, I think this is as good as we can get it. :-)

PR #17, opened 2021-04-20
Rewrite/expansion of Solr indexing.

demiankatz:
TODO
- [x] Add ability to define the top of the hierarchy
- [x] Add missing hierarchy/sequence fields
  - [x] hierarchy_browse
  - [x] hierarchy_top_id
  - [x] hierarchy_top_title
  - [x] hierarchy_parent_id
  - [x] hierarchy_parent_title
  - [x] hierarchy_first_parent_id_str
  - [x] hierarchy_all_parents_str_mv
  - [x] is_hierarchy_id
  - [x] is_hierarchy_title
  - [x] sequence_[pid]_str
- [x] Add TODOs for missing sequence fields that are not yet easy to implement
  - [x] has_order_str
  - [x] hierarchy_sequence
  - [x] hierarchy_sequence_sort_str
- [x] Add relsext.* indexing.
- [x] Add dc.* indexing.
- [x] Improve accuracy/comprehensiveness of Dublin Core rules.


demiankatz:
I think this is ready to merge. I've implemented indexing for the most critical hierarchy fields, and added TODOs for some secondary things (like fixed sequencing) which I don't have good data to test with right now. I think we should migrate some real data to create a better test set, and then we can polish this further.

I'm seeing some differences between the behavior of my code and the real index (mainly with how things in the holding area are handled) so adjustments may be needed, but I think this is going to do what we need for the majority of cases. We'll just have to be more aware of the edge cases when we begin porting the Orbeon logic, because I have less familiarity with how that uses Solr.

We can discuss further tomorrow!

demiankatz:
I was able to implement indexing for one more field from the RELS-EXT, but now I seem to have hit a wall related to data that's stored in the FOXML and appears to possibly not be getting migrated by the migration utility. I'm not 100% sure we actually need this, but we should definitely investigate further to be sure there's not a gap in migration functionality.

demiankatz:
Okay, now I'm done for real. While reviewing the original XSLT, I discovered there are several wildcard fields that get populated that we missed in our initial field inventory... in order to implement those, I had to refactor code significantly... and that led to a total rewrite of the Dublin Core indexing, which included correcting some problems, adding some missing details, and modernizing some things for compatibility with newer VuFind releases.

Let's discuss, and hopefully merge, this tomorrow.

Next steps:
- Continue populating more fields; in some cases, I think we may need to revisit the migration tool to figure out where data is going (or whether it's getting lost in transition)
- Actually start pushing data to Solr; I think we have enough here that we can start to build out VuFind customizations once the index is populated
- Migrate real data; there are a lot of fields I can't really test because of limitations in the sample repository.

demiankatz:
One more last-minute change: some of these fields were named according to GSearch conventions, and would not work with VuFind's standard schema. I added field suffixes to make them VuFind-compatible. We'll have to keep this in mind when porting legacy code, but since we have to rewrite everything anyway, it seems like a small price to pay to avoid needing a custom schema.

PR #22, opened 2021-04-23
Fill in more indexing rules.

demiankatz:
All of our previous indexing efforts were focused on data streams, but this adds support for information that originally lived in the FOXML but which the migration script moves into the Fedora 6 native linked data storage.

Note that I'm working with this data using XML+RDF because it allows me to recycle code originally written for the RELS-EXT stream. We could probably make this run faster by using JSON instead, but that requires more work to code since in this instance, XPath actually makes the data extraction really easy.


demiankatz:
Okay, I think this is done now!

demiankatz:
I've done some more work on this, and I think it's now ready to be merged (with approval from @Geoffsc and/or @crhallberg). The only fields which are not yet being indexed are first_indexed and last_indexed, which require integration with a VuFind web service that I haven't set up yet -- I'll deal with that as a separate PR. There's also some cleanup in terms of making more things configurable, but again, I'd prefer to deal with that as a separate PR, since this one is already very large. The most recent work I've done adds full text extraction using Tika, and significantly refactors the Fedora class to improve/clarify caching and to eliminate use of `any` to make Typescript happier.

Let me know what you think!

PR #24, opened 2021-04-27
Job metadata

Geoffsc:


api/src/models/JobMetadata.ts
```diff
@@ -1,12 +1,16 @@
 import PageOrder from './PageOrder';
 import DocumentOrder from './DocumentOrder';
+import AudioOrder from './AudioOrder';
 import Job from './Job';
 
 class JobMetadata {
     job: Job;
+    _filename: string;
```
demiankatz, line 8: This property currently does not appear to be used. It looks like the Ruby code uses the filename property in both the constructor and the save method. Maybe you should make some adjustments here to match.

api/src/models/JobMetadata.ts
```diff
@@ -1,12 +1,16 @@
...
     _order: PageOrder = null;
-    _documents: DocumentOrder = null
-    published: boolean = false;
+    _documents: DocumentOrder = null;
+    _audio: AudioOrder = null;
+    _published: Boolean = false;
+    published: Boolean = false;
```
demiankatz, line 12: Why do you have _published AND published here?

I think the general rule is to use the underscore for private/internal properties, and no underscore for public ones. I think in this case, you can just make it a public property, as there is no need to associate logic with it. (The method in the Ruby version of the code is just there to initialize the variable to false, but having a default value in the declaration here does the job just as well, if not better).

api/src/models/JobMetadata.ts
```diff
@@ -25,6 +29,22 @@ class JobMetadata {
         }
     }
 
+    ingestLockfile(job) {
```
demiankatz: This method shouldn't need to take a job parameter; I believe it can use this.job instead, since the job was passed to the constructor and saved internally.

api/src/models/JobMetadata.ts
```diff
@@ -25,6 +29,22 @@ class JobMetadata {
...
+    get raw() {
+        return {
+            order: this.order.raw,
+            published: this.published === true
+        };
+    }
+
+    set raw(data) {
+        this._order = PageOrder.fromRaw(data);
+        this._published = true;
```
demiankatz: The published property is already a boolean, so I think you can just stick it here directly; no need for the extra `=== true`.

api/src/models/JobMetadata.ts
```diff
@@ -83,30 +99,33 @@ class JobMetadata {
...
     }
 
-    get raw() {
-        return {
-            order: this.order.raw,
-            published: this.published  
-        };
+    get audio() {
+        if (this._audio === null) {
+            this._audio = AudioOrder.fromJob(this._audio);
```
demiankatz: I think this should be `fromJob(this.job)`, not `fromJob(this._audio)`.

demiankatz: Looks like the documents() method above has the same bug (or a different variation thereof).

demiankatz (changes requested):
Thanks for the progress, @Geoffsc! I've left a few comments.

With the save method written, it may actually be possible to hit the save button after numbering some pages, then reopen the job to resume your work. I'd recommend doing some end to end testing to see if that's working yet. Another milestone!

api/src/models/JobMetadata.ts
```diff
@@ -83,30 +98,33 @@ class JobMetadata {
...
-    set raw(data) {
-        //TODO: set raw data
+    set audio(data) {
+        this._audio = AudioOrder.fromRaw(data);
     }
 
     save() {
-
+        let fs = require('fs');
+        this._filename = this.job.dir + '/job.json'
```
demiankatz: The constructor is building this same filename string in a local variable. Since you can guarantee that the constructor is always called first, I'd suggest moving this assignment into the constructor, using `this._filename` there in place of the local `filename` variable... and then you don't need to recalculate the filename here, since it will already be set.

demiankatz (changes requested):

api/src/models/JobMetadata.ts
Geoffsc: Updated, thanks!

demiankatz: I see that you moved the assignment to `this._filename` into the constructor, but you didn't replace the redundant local `filename` variable. I'd recommend doing that to simplify everything.

PR #27, opened 2021-05-03
JobMetadata updates

Geoffsc:


package.json
```diff
@@ -13,19 +13,15 @@
     "api:wsl": "nodemon --legacy-watch api/bin/www",
     "api:build": "cd api && tsc",
     "api:watch": "cd api && tsc --watch",
-
```
demiankatz: @crhallberg put this whitespace here for clarity, but I anticipate that it's going to get auto-reverted and accidentally committed endlessly. @crhallberg, should we just bite the bullet and let it get flattened out so we don't spend time constantly fixing it? Or do you think it's worth the effort?

api/src/models/ImageFile.ts
```diff
@@ -71,11 +72,9 @@ class ImageFile {
     }
 
     async ocr(): Promise<void> {
-        //TODO: update the following (derivativepath requires size now)
```
demiankatz, line 74: Is this TODO really done? Just curious why the only changes in this file are the removal of comments and reorganization of imports... is this intentional or a merge artifact?

api/src/models/JobMetadata.ts
```diff
@@ -58,31 +60,61 @@ class JobMetadata {
     }
 
     get derivativeStatus(): Record<string, unknown> {
-        // TODO: populate with real data
+        let lockfileExists: boolean = fs.existsSync(this.derivativeLockfile);
```
demiankatz: Should this be const since it is never changed?

api/src/models/JobMetadata.ts
```diff
@@ -58,31 +60,61 @@ class JobMetadata {
...
         const status = {
             expected: 10,
```
demiankatz: The 10 here was a fake value we made up just to see something in the UI. This should actually get initialized to 0 now that we're doing work for real.

api/src/models/JobMetadata.ts
```diff
@@ -58,31 +60,61 @@ class JobMetadata {
...
             processed: 0,
             building: false,
+            lockfileExists
```
demiankatz: lockfileExists should be the value of the building key; here it's just getting added as a free-floating element in the object.

api/src/models/JobMetadata.ts
```diff
@@ -58,31 +60,61 @@ class JobMetadata {
...
+        });
         return status;
     }
 
     get uploadTime(): number {
-        // TODO: populate with real data
-        return 0;
-    }
+        this.order.pages.forEach(element => {
+            let getFileUpdatedDate = (path: string = this.job.dir + "/" + this.page.filename) => {
```
demiankatz: I don't think this is doing what it needs to do. The original code is looking for the newest file within the job, so it iterates through files and keeps track of the newest time it has encountered so far. If it finds nothing at all, it falls back to the age of the directory itself. The forEach code doesn't seem to be doing any comparison to find a new date, and nothing seems to be manipulating the this.uploadTime property that the function ultimately returns (is that even defined anywhere?). I'm also not sure why there is a callback that returns a call to a callback here...

api/src/models/JobMetadata.ts
```diff
@@ -58,31 +60,61 @@ class JobMetadata {
...
 
-    get ingestInfo(): string {
-        // TODO: populate with real data
-        return "";
+    get ingestInfo() {
+        let readLastLines = require('read-last-lines');
+        let logfile: string = this.job.dir + "/ingest.log";
+        if (fs.existsSync(logfile)) {
+            return readLastLines.read(logfile, 1)
+                .then((lines) => console.log(lines));
```
demiankatz: I think this is probably going to send the line to the console instead of returning it from the function. You may need to make this async and use an await here.

demiankatz (changes requested):
Thanks, @Geoffsc! I've left a review. A lot of this looks like good progress, but there are a couple places that I think are going to need more work (especially uploadTime()). I'm also pretty sure the linter is going to make you change a bunch of `let`'s to `const`'s and boost your include to the top of the file, so be ready for that. :-)

package.json
crhallberg: Yeah we don't need to fight it

api/src/models/JobMetadata.ts
crhallberg: Do we want a lint action on PRs?

demiankatz: If there's an easy way to wire it up, it wouldn't hurt.

demiankatz: Not sure if we want to do CI stuff in Geoff's personal repo, though; maybe we should think about moving this under FalveyLibraryTechnology soon.

api/src/models/JobMetadata.ts
```diff
@@ -58,31 +61,67 @@ class JobMetadata {
...
+                }
+            });
+        });
         return status;
     }
 
     get uploadTime(): number {
-        // TODO: populate with real data
-        return 0;
+        let mtime = 2000;
```
demiankatz: It might be nice to use `const undefined_time = 2000;` above this, and then replace both hard-coded `2000`'s with `undefined_time`, just to eliminate a "magic number" from the code.

demiankatz (commented):
This is looking great to me, @Geoffsc -- just one small suggestion to make this slightly more in line with the original Ruby code. Once that's done, I think this is ready to merge.

I think the next step after this is to get job saving working, which requires us to figure out where to put the validation logic. We can discuss that tomorrow. I imagine if you search the code for TODOs you might find some other low-hanging fruit to work on in the meantime.

PR #28, opened 2021-05-03
Add Lint Action to PRs Maybe

crhallberg:
- [x] Run eslint on all PRs
- [x] Auto fix prettier format issues
- [ ] Do we also want to lint all repository pushes?


crhallberg:
I have no idea if this works and I'm having a hard time finding decent docs or guides for this so maybe forget it.

demiankatz:
I'm guessing that we don't want to actually run `--fix` here; we probably want to do the regular lint so it dumps out a report of things that need work. I'm not sure how you actually get the job to run, or how you control the status of the PR based on the return value of the command you run. Is it possible that @Geoffsc needs to turn something on before GitHub will even look at this .yml file? I'm guessing it would be kind of a security risk if anybody could submit a PR to a random project and start running tasks as a result.

crhallberg:
I agree. I kept in `--fix` because I wanted to see if it would automatically add commits to the PR.

crhallberg:
-gasp- it worked!

crhallberg:
So [as you can see](https://github.com/Geoffsc/VuDLPrepJS/pull/28/checks#step:8:13) eslint will not fail a PR when only a warning exists. I found a setting to fail on warnings for webpack but not for eslint itself yet.

demiankatz:
Hooray! So now the question is, why is that particular file failing here but not when I run the linter manually on my system? Was that introduced recently, or am I doing something wrong?

crhallberg:
I introduced a bug as a test and demonstration 😉

PR #37, opened 2021-05-23
Improve .ini/config handling.

demiankatz:
- Use dependency injection to pass .ini contents into PrivateConfig (so we load the file only once, instead of every time a setting is requested).
- Eliminate config methods in model classes (unneeded, since we can use Config.getInstance() directly).
- Use typecasting, so we don't have to suppress eslint warnings about use of any.

### DIFF
--- a/api/src/jobs/Ingest.ts
+++ b/api/src/jobs/Ingest.ts
@@ -234,7 +234,7 @@ class IngestProcessor {
     }
 
     moveDirectory() {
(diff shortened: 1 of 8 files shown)


demiankatz:
I would also like to make all of the methods in PrivateConfig getters, so we can access the values without using extra parentheses... but since we have a bunch of pull requests in progress right now, that would create some conflicts and problems, so we can do that later after other outstanding work has been merged.

PR #38, opened 2021-05-26
Completed audio derivative generation

Geoffsc:
Fixed a bunch of bugs, tested and the mp3 and ogg derivatives are getting generated now. I had to switch from the async exec to the synchronous exec because the method was attempting to return the derivative before it was done being generated. TODO: update the derivative filenames (currently it's appending the new extension i.e. "vudl_567755_MASTER.flac.mp3", "vudl_567755_MASTER.flac.ogg"

### DIFF
--- a/api/src/models/AudioFile.ts
+++ b/api/src/models/AudioFile.ts
@@ -13,11 +13,6 @@ class AudioFile {
         this.dir = dir;
     }
 
-    ffmpeg(): string {
-        const ffmpeg_path = Config.getInstance().ffmpegPath;
-        return ffmpeg_path;
-    }
-
     derivative(extension: string): string {
         const deriv = this.derivativePath(extension);
         if (!fs.existsSync(deriv)) {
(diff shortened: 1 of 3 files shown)


api/src/models/AudioFile.ts
```diff
@@ -14,14 +13,30 @@ class AudioFile {
         this.dir = dir;
     }
 
+    ffmpeg(): string {
```
demiankatz: Is there value in having this ffmpeg method at all? Since this is only used in one place, maybe just moving `const ffmpeg_path = Config.getInstance().ffmpegPath;` down into the derivative method would be just as good.

If you do want to keep this (which I wouldn't object to -- just posing a question), I would suggest making this a getter, since it doesn't really need to be a method, and then you can take some parentheses off down below.

api/src/models/AudioFile.ts
```diff
@@ -33,7 +48,7 @@ class AudioFile {
     }
 
     derivativePath(extension = "flac"): string {
-        const filename = path.basename(this.filename);
+        const filename = path.basename(this.filename, ".*");
```
demiankatz: The Node basename function doesn't support ".*" -- it's a little more work, but you can copy my solution from here:

https://github.com/Geoffsc/VuDLPrepJS/blob/dev/api/src/models/ImageFile.ts#L61-L66

demiankatz (commented):
Looking great, @Geoffsc -- just left one comment for your consideration, and another with a hint to help solve your remaining TODO!

demiankatz:
@Geoffsc, based on our conversation today, I went ahead and got this to what I think is a mergeable state. The only minor concern I had is that when I ran it, I discovered that it was broken due to the missing slash in the glob in the AudioOrder model. I know that you fixed that, so I'm not sure why it wasn't here. Is it possible there are any other changes that didn't get pushed somehow?

(I know that we also changed the queue:dev script in package.json, but I don't think we need to push that here at this point -- @crhallberg came up with a slightly better version of my proposed fix elsewhere).

PR #45, opened 2021-06-12
Add main menu and Solr tool to React app.

demiankatz:
- Also changes the base url to allow access to routes outside of /api


demiankatz:
As I mentioned a while back, I thought it would be helpful to have a proper main menu. I also built a tool wrapping the Solr indexing functionality as an example of a secondary menu item, which required some minor adjustments to the ajax helper. This could use some aesthetic improvements, but it is functional for now. Once we get this merged, I can add some new tools (like the PDF generator for existing objects).

PR #48, opened 2021-06-13
Begin work on actually adding data to F6.

demiankatz:
This is WORK IN PROGRESS on the real Fedora 6 ingest.

TODO:
- [x] Merge #46, since it is a pre-requisite for the work here.
- [x] Get container ingest working
- [x] Get datastream ingest working
- [x] Full end-to-end testing


demiankatz:
As of today, this is totally non-working, early work in progress. I just did enough today to confirm that I can run F6 inside my VM and create data within it using the API. This code is not yet doing anything correctly, but it does include some refactoring to simplify code, remove F3-specific details that no longer matter, and make more things asynchronous where they need to be. I'm hoping we can work together on actually making this work as part of a future call.

demiankatz:
I think I've done it -- all data is now loading into Fedora 6! I just need to spend some time testing next week to be sure it's really working.

demiankatz:
Now that I have successfully completed end-to-end testing of the ingest, I think we can go ahead and merge this if no one objects. There are still some TODOs here (FITS support, moving the directories to the "processed" area at the end of the ingest), but I'd love to get this merged sooner rather than later. I can work on those other things in separate PRs.

PR #52, opened 2021-06-24
sanitization updates

Geoffsc:


api/src/routes/api.ts
```diff
@@ -6,11 +6,13 @@ import Config from "../models/Config";
 import Job from "../models/Job";
 
 const router = express.Router();
+const { body, validationResult } = require("express-validator");
```
demiankatz: I don't think this is getting used at all.

api/src/routes/api.ts
```diff
@@ -6,11 +6,13 @@ import Config from "../models/Config";
...
 import { setupPassport, requireToken } from "./auth";
+import fs = require("fs");
 setupPassport(router);
 
 function getJobFromRequest(req): Job {
-    // TODO: sanitize parameters!
+    //TODO make sure Job exists
```
demiankatz: It should be possible to do this very similarly to the check in the /:category route. You just need to see if the job directory (`holdingArea() + req.params.category + "/" + req.params.job`) actually exists. If it doesn't, you need to figure out how to trigger a 404 error. Perhaps you could return null from this function, and then you have to check the return value in all the places where this gets called. Maybe there's an easier way, like creating an "addJobToRequest" callback that you add to the chain in the route, that creates a this.request.job based on the other request parameters -- then the final callback in each chain could use this.request.job instead of calling getJobFromRequest, and this function could just return the 404 and break the chain if the file is not found.

demiankatz (commented):
Thanks, @Geoffsc -- just a couple of comments. Have you had a chance to thoroughly test this? It's particularly worthwhile to test the delete button in the paginator, since that code was in a totally broken state and thus has not been tested (though be warned that it will actually delete files if it works, so make a backup of your test jobs). I can help with testing next week as needed; just not sure how much time I should put into this beyond what you've already done.

api/src/routes/api.ts
Geoffsc: Fixed!

api/src/routes/api.ts
Geoffsc: Updated!

api/package.json
```diff
@@ -16,6 +16,7 @@
     "ejs": "^3.1.6",
     "express": "~4.16.1",
     "express-session": "^1.17.1",
+    "express-validator": "^6.12.0",
```
demiankatz: I don't think we ended up using this; can we remove it?

api/package.json
```diff
@@ -27,6 +28,7 @@
     "nanoid": "^3.1.23",
     "needle": "^2.6.0",
     "node-tesseract-ocr": "^2.0.0",
+    "npm": "^7.18.1",
```
demiankatz: Why are we adding NPM here?

api/src/routes/api.ts
```diff
@@ -7,10 +7,13 @@ import Job from "../models/Job";
 
 const router = express.Router();
 import { setupPassport, requireToken } from "./auth";
+import fs = require("fs");
 setupPassport(router);
 
 function getJobFromRequest(req): Job {
-    // TODO: sanitize parameters!
+    if (!fs.existsSync(holdingArea() + req.params.category + "/" + req.params.job)) {
+        return null; //res.status(404).json({ error: "Not Found"});
```
demiankatz: All of the code calling getJobFromRequest needs to be able to handle the possibility that it will return null, and return a 404 when it does. Right now it looks like everything just assumes this will work, which is a problem.

demiankatz (changes requested):

api/package.json
Geoffsc: fixed!

api/package.json
Geoffsc: not sure how that happened, fixed!

PR #56, opened 2021-07-05
Add PDF generator job.

demiankatz:
This pull request adds the functionality to generate PDFs for existing objects in the repository that do not already have them.

Note that this also adds authentication and parameter sanitization to the messenger routes (which was previously missing) and refactors/revises the sanitizeParameters function to make that possible.

This is the last critical piece of functionality from the Ruby VuDLPrep code that needed to be ported over here. (Unless you count the OCR regenerator job, which might also be worth porting, though we don't currently have an active use case for it).

TODO
- [x] Merge #48, #54 and #57, because the work here depends on all of those
- [x] Finish the outstanding work on FITS-related Solr indexing, because we cannot index PDFs correctly without FITS metadata, which makes some parts of this work (such as actually viewing the PDF in VuFind, and detecting whether PDFs already exist) impossible to test.


demiankatz:
@Geoffsc, I have just merged the dev branch into this PR, made some minor adjustments, and tested everything. I believe this is ready to merge with your approval.

PR #60, opened 2021-07-22
Add command-line tool to run ingests.

demiankatz:
TODO
- [x] Fix hanging problem

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,8 @@ After a few moments, a new tab should automatically open in your browser pointin
 | queue:dev | start api worker listener, restart on changes |
 | queue:wsl | start api worker listener, restart on changes (use if on WSL or `api:dev` isn't working) |
 | | |
+| ingest | add all published jobs to the ingest queue |
+| | |
 | backend | start both api and worker queue servers |
 | backend:dev | start both api and worker queue servers, restart on changes |
 | backend:wsl | start both api and worker queue servers, restart on changes (use if on WSL or `api:dev` isn't working) |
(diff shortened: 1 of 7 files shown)


demiankatz:
This PR adds a small command-line tool to ingest published jobs, for use in cron, etc. This is the equivalent of [the old Rake task](https://github.com/FalveyLibraryTechnology/VuDLPrep/blob/master/lib/tasks/ingest.rake).

I did a bit of refactoring around loading of the holding area configuration. We had a method in the api.js route file which normalized the holding area path to ensure a trailing slash, but it makes more sense to do that normalization inside the configuration object, so we can reliably get the same result everywhere. Since I needed the path in order to build this tool, I made those changes.

There is one outstanding problem which I don't understand: when jobs actually need to be published, the tool hangs indefinitely. If there are no jobs to be published, it terminates as expected. I have some suspicion that this is another issue related to using the shared file system in VirtualBox, but I haven't had time to test using only local files yet. I will do that when time permits; if that doesn't address the issue, we may have more digging to do. There's not actually very much code executing here, so there are a limited number of possible explanations.

demiankatz:
I've narrowed down the cause of the hanging -- it's related to bullmq, not the file system. This research revealed that the method to add things to queues is actually asynchronous, and the examples suggest we should be using it with `await`. I've added commit 03dca73 to address that detail. However, it does not solve the problem. I'm nearly out of time but will continue to investigate as time permits next week.

demiankatz:
Ha-ha! Got it with moments to spare! The problem is that you need to close Queue objects when you are done with them. I've added the missing close calls in 581158d, and now this is working as expected.

PR #61, opened 2021-08-04
Dont forget about the children

Geoffsc:


api/views/login-test.ejs
```diff
@@ -1,4 +1,4 @@
-<form action="/api/login">
+<form action="/api/ingest/login">
```
demiankatz: Rather than having the login routes under /api/ingest, it probably makes more sense to have them in the root of /api... I'm just not entirely sure how best to set that up, or exactly what role the setupPassport function has. Do we need to run setupPassport on all of our routers, or is it a one-time thing that should be called from app.js somehow? We should talk to @crhallberg about this before merging this work.

client/src/AjaxHelper.js
```diff
@@ -8,20 +8,20 @@ class AjaxHelperInstance {
         this.handle4xx = this.handle4xx.bind(this);
     }
 
-    get apiUrl() {
-        return this.url + "/api";
+    get ingestApiUrl() {
```
demiankatz, line 19: We should also talk to @crhallberg about how we'll manage URLs when we have separate /edit and /ingest routes. Maybe we still need a `get apiUrl()` method that returns simply `this.url`.

demiankatz (commented):
This is looking pretty good to me, but I think we should discuss our route rearranging as a group on tomorrow's call before merging it, in case we refine anything. See specific notes below.

PR #69, opened 2021-08-06
Use constructor-injection to reduce coupling.

demiankatz:
I had a brainstorm last night while thinking about our conversation about testing.

As written, the code is hard to test because it is too tightly coupled -- the reliance on the Config singleton makes it hard to override things for testing purposes, and we have a lot of places where objects instantiate other objects inline, which makes it impossible to use test doubles.

It occurred to me that this is pretty easy to avoid by making some minor changes, following these rules:

1.) We should NEVER call a singleton directly; we should only inject it through a constructor.
2.) We should NEVER instantiate a dependency directly; we should only inject it through a constructor.
3.) For objects that can operate as a singleton (i.e. the Config and most of the services) we should create a static getInstance() method which performs default dependency injection by using appropriate singletons and factories.
4.) For objects that have multiple instances, we should create a static build() factory method, which does more or less the same thing as getInstance, but accepts additional parameters to use for building the object.

This way, we have an easy default way to get all of our objects composed together the way we like, but these relationships are not hard-coded into the classes, so it's much easier to use the code in non-standard ways like for testing purposes.

I haven't completely implemented this idea, because I want to make sure we all agree on this approach before I take that much time. However, I did refactor the SolrIndexer class and all of its dependencies as an example. (I also combined the Config and PrivateConfig classes into a single class -- this eliminates the enforced singleton nature of Config, which I think is desirable for testability). If you like this, I'll do the rest of the work. Thanks to the Typescript compiler, it's really easy to do this refactoring safely, since the compiler catches you if you miss any details.

TODO
- [x] Discuss
- [x] Finish implementing proposal
- [ ] Decide on a policy for route callbacks -- is it okay to call singletons here, or should we make these more testable as well?


demiankatz:
Okay, I've taken this a bit further, and in my opinion it's now in a merge-ready state (though I'd like to hear @crhallberg's thoughts first, in case he has any suggested changes to the conventions). More work could be done in a few places -- there may be some cases where using the build() factory method is less desirable than using a direct constructor call, if the object being created has the same dependencies as the object creating it, and they can be directly passed down (I didn't check for this situation); it may also be worth discussing whether routes should be generated in a more flexible way, as noted in the TODO list above. But I think we could merge this as incremental progress, if we're all happy with it.

PR #73, opened 2021-08-24
Add unit testing and function component template

diboy2:
TODO
- [x] Write unit tests
- [x] Move session storage 
- [x] Incorporate unit tests into git check
- [x] Clean up code
- [x] Add ability to check props on mocked react components
- [x] Add ability to publish code coverage moved to task. Too much configuration

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -17,8 +17,18 @@
   "scripts": {
     "start": "react-scripts start",
     "build": "react-scripts build",
-    "test": "react-scripts test",
+    "test": "react-scripts test --coverage --watchAll=false",
     "eject": "react-scripts eject"
+  },
+  "jest": {
+    "collectCoverageFrom": [
+      "src/**/*.{js,jsx,ts,tsx}",
+      "!<rootDir>/node_modules/"
+    ],
+    "coverageReporters": ["text"],
+    "snapshotSerializers": [
+      "enzyme-to-json/serializer"
+    ]
   },
   "eslintConfig": {
     "extends": "react-app"
@@ -34,5 +44,10 @@
       "last 1 firefox version",
       "last 1 safari version"
     ]
+  },
+  "devDependencies": {
+    "enzyme": "^3.11.0",
+    "enzyme-adapter-react-16": "^1.15.6",
+    "enzyme-to-json": "^3.6.2"
   }
 }
--- a/client/src/Category.test.js
+++ b/client/src/Category.test.js
@@ -1,0 +1,64 @@
+import React from "react";
+import { beforeEach, describe, expect, it, jest } from "@jest/globals";
+import { render, shallow, mount } from "enzyme";
+import toJson from "enzyme-to-json";
+import Category from "./Category";
+
+const mockjobList = jest.fn();
+jest.mock(
+    "./JobList",
+    () =>
+        function JobList(props) {
+            mockjobList(props);
+            return <mock-JobList />;
+        }
+);
+
+describe("Category", () => {
+    let props;
+
+    beforeEach(() => {
+        props = {
+            data: {
+                category: "testCategory",
+                jobs: ["testJobs"],
+            },
+        };
+        Object.defineProperty(window, "sessionStorage", {
+            value: {
+                getItem: jest.fn(() => null),
+                setItem: jest.fn(() => null),
+            },
+            writable: true,
+        });
+    });
+
+    it("renders", () => {
+        const wrapper = shallow(<Category {...props} />);
+        expect(toJson(wrapper)).toMatchSnapshot();
+    });
+
+    it("renders no jobs", () => {
+        props.data.jobs = [];
+        const wrapper = render(<Category {...props} />);
+        expect(wrapper.has("no jobs")).toBeTruthy();
+    });
+
+    it("calls setItem on toggle", () => {
+        const component = mount(<Category {...props} />);
+        expect(sessionStorage.setItem).toHaveBeenCalledWith("open-testCategory", "false");
+        expect(mockjobList).not.toHaveBeenCalled();
+
+        component.find("button").simulate("click");
+
+        expect(sessionStorage.setItem).toHaveBeenCalledWith("open-testCategory", "true");
+        expect(mockjobList).toHaveBeenCalledWith(
+            expect.objectContaining({
+                category: props.data.category,
+                data: props.data.jobs,
+            })
+        );
+
+        component.unmount();
+    });
+});
--- a/client/src/sessionStorage.js
+++ b/client/src/sessionStorage.js
@@ -1,0 +1,22 @@
+import { useState, useEffect } from "react";
+
+const getSessionStorage = (storageKey, initialStorageValue) => {
+    if (typeof window.sessionStorage === "undefined" || !window.sessionStorage.getItem(storageKey)) {
+        return initialStorageValue;
+    }
+    return JSON.parse(window.sessionStorage.getItem(storageKey));
+};
+
+const useSessionStorage = (storageKey, initialStorageValue) => {
+    const [storageValue, setStorageValue] = useState(getSessionStorage(storageKey, initialStorageValue));
+
+    useEffect(() => {
+        if (typeof window.sessionStorage !== "undefined") {
+            sessionStorage.setItem(storageKey, JSON.stringify(storageValue));
+        }
+    }, [storageKey, storageValue]);
+
+    return [storageValue, setStorageValue];
+};
+
+export { getSessionStorage, useSessionStorage };
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "client:build": "cd client && npm run build",
     "client:lint": "eslint client/src/ --ext .ts,.js,.jsx",
     "client:setup": "cd client && npm ci",
+    "client:test": "cd client && npm test",
     "queue": "node api/worker-start.js",
     "queue:dev": "nodemon --watch 'api/vudl.ini' --watch 'api/**/*.js' api/worker-start.js",
     "queue:wsl": "nodemon --legacy-watch api/worker-start.js",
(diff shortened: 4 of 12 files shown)


demiankatz:
@diboy2, this looks really good to me. I haven't actually done hands-on testing yet, but the refactoring all makes sense. (I did have to look up how state hooks work, since my React knowledge was rusty, but I get it now). :-) When you work on testing tomorrow, do you want to push your test work to this same branch, and then we can merge the refactoring and corresponding tests all at once?

diboy2:
@demiankatz I'm planning to push my unit test work on the same branch. My approach is usually to add unit tests to my work. That usually adds a little more overhead.

demiankatz:
Excellent, I definitely support that practice! You might also find it helpful to use the checkbox feature of pull requests to mark the status of work in progress. For example, if I were working on this PR, I'd probably edit the top-level comment to something like:

<pre>
TODO
- [ ] Write unit test
</pre>

...and then that's a signal to the rest of us that there's work in progress, and it shouldn't be merged yet. (You can also use the "draft" status feature that was recently added on GitHub... but I like the checkbox approach since it's more informative).

client/src/sessionStorage.js
```diff
@@ -0,0 +1,22 @@
+import { useState, useEffect } from "react";
+
+const getSessionStorage = (storageKey, initialStorageValue) => {
+    if (typeof window.sessionStorage !== "undefined" || !window.sessionStorage.get(storageKey)) {
```
demiankatz: Should this be `===` rather than `!==` -- that is, return the default value if session storage is NOT defined, rather than if it is?

diboy2: yup. I saw that while I added unit tests.

package.json
```diff
@@ -18,6 +18,7 @@
     "client:build": "cd client && npm run build",
     "client:lint": "eslint client/src/ --ext .ts,.js,.jsx",
     "client:setup": "cd client && npm ci",
+    "client:test": "cd client && npm test",
```
demiankatz, line 21: We should add this new option to the npm list in the README, just to keep the docs in sync with the configuration.

client/src/Category.test.js
```diff
@@ -0,0 +1,49 @@
...
+import toJson from "enzyme-to-json";
+import Category from "./Category";
+
+jest.mock("./JobList", () => () => "JobList");
+
+describe("Category", () => {
+    const props = {
+        data: {
+            category: "test",
+            jobs: ["test"],
```
demiankatz: It might be a good idea to use "testCategory" and "testJob" here, just to make it more clear which test value is being referred to in which context.

client/src/Category.test.js
```diff
@@ -0,0 +1,49 @@
...
+
+    it("renders", () => {
+        const wrapper = shallow(<Category {...props} />);
+        expect(toJson(wrapper)).toMatchSnapshot();
+    });
+
+    it("renders no jobs", () => {
+        props.data.jobs = [];
+        const wrapper = shallow(<Category {...props} />);
+        expect(wrapper.contains("no jobs")).toBeTruth();
```
demiankatz: Is `toBeTruth` a typo here?

demiankatz:
@diboy2, thanks again for the really excellent work here -- this is looking great! I left a few comments based on a fairly cursory scan of your work. I realize some things are still in progress, but I leave these just in case they save you any time or trouble. We'll talk about it all in more detail tomorrow!

client/src/Category.test.js
diboy2: yes made change

demiankatz:
Thanks for the further updates, @diboy2. Two further suggestions (which we can discuss at today's meeting):

1.) Right now, continuous integration is failing because Prettier doesn't like the __snapshots__ directory. I think we should either move the snapshots outside of src, or else we need to adjust the Prettier configuration so it doesn't get upset by them.

2.) Speaking of continuous integration, now that we have tests, we should incorporate them into the CI process.

demiankatz:
Oh, one other thing for consideration (though we certainly don't need to address it in this particular pull request): as we refactor code to make things more granular and easier to test, we're likely to have a proliferation of files. Do we need to talk about file arrangement to make it easier to navigate the code? Do we want to separate helpers from components? Do we want to organize components by function (e.g. separate stand-alone utilities like PdfGenerator and SolrIndexer from the paginator functionality and the forthcoming editor)?

diboy2:
@demiankatz I had the same thought about file organization. I think it would become too unwieldy if we had all the components to in one folder, so I'm proposing we have a task to have a better file structure.

demiankatz:
Good idea, @diboy2 -- I've added a new "Improve React code" card to our board, with a few sub-tasks: file structure, modernization (which you're currently working on), and investigating alternatives to create-react-app (as you know, I'm not very experienced with current React practices, but it seems that using create-react-app brings in a lot of unnecessary dependencies and causes many NPM security audit warnings, and I wonder if there's a simpler and cleaner way to get our stack running).

diboy2:
For the audit issues, running npm audit and keeping those packages updated
would resolve of the majority issues.

Adrian Cagaanan

demiankatz:
@diboy2, that's good to hear -- if we just need to audit more frequently as part of our workflow and that solves our problems, then I'm happy with that solution. Something else we can discuss today!

demiankatz:
Just a quick update to say that I did some hands-on testing of this in my own environment, and it seems to be working as expected for me. I ran into a weird issue where after refreshing the paginator page, I got an inexplicable CORS error that prevented category information from loading, but this was after testing with multiple tabs open at the same time, so I suspect I may have gotten my browser into a confused state. Refreshing a second time fixed it. There is quite possibly a bug somewhere, but I doubt it's related to the refactoring in this branch.

I've also opened #74 as a place to work through `npm audit` results.

package.json
diboy2: Made change

client/src/Category.test.js
diboy2: Made change

client/src/Category.test.js
```diff
@@ -0,0 +1,60 @@
...
+
+    it("renders", () => {
+        const wrapper = shallow(<Category {...props} />);
+        expect(toJson(wrapper)).toMatchSnapshot();
+    });
+
+    it("renders no jobs", () => {
+        props.data.jobs = [];
+        const wrapper = shallow(<Category {...props} />);
+        expect(wrapper.contains("no jobs")).toBeFalsy();
```
demiankatz: Apologies if I'm just misunderstanding or misreading something, but why are we expecting this to be falsy? Don't we expect that the wrapper SHOULD contain "no jobs" in this case, rather than NOT CONTAIN that string?

client/src/Category.test.js
```diff
@@ -0,0 +1,60 @@
...
+    });
+
+    it("calls setItem on toggle", () => {
+        const component = mount(<Category {...props} />);
+        expect(sessionStorage.setItem).toHaveBeenCalledWith("open-testCategory", "false");
+        expect(component.contains("JobList")).toBeFalsy();
```
demiankatz: Here's another toBeFalsy call that I don't fully understand.

client/package.json
```diff
@@ -17,9 +17,20 @@
...
     "build": "react-scripts build",
-    "test": "react-scripts test",
+    "test": "react-scripts test --coverage --watchAll=false",
     "eject": "react-scripts eject"
   },
+  "jest": {
+    "collectCoverageFrom": [
+      "src/**/*.{js,jsx,ts,tsx}",
+      "!<rootDir>/node_modules/",
+      "!<rootDir>/path/to/dir/"
```
demiankatz: Is "/path/to/dir/" an example that we don't really need to retain here?

package.json
```diff
@@ -27,10 +28,11 @@
     "backend:wsl": "concurrently npm:api:wsl npm:queue:wsl",
     "build": "concurrently npm:api:build npm:client:build",
     "dev": "concurrently -c 'cyan,green,gray' npm:client npm:api:dev npm:queue:dev",
-    "format": "prettier --check --write api/*.js api/src/** client/src/**",
+    "format": "prettier --check --write api/*.js api/src/** client/src/*.{js,jsx}",
     "lint": "eslint api/*.js api/src/ client/src/ --ext .ts,.js,.jsx --max-warnings 0",
     "setup": "concurrently npm:api:setup npm:client:setup",
     "start": "concurrently -c 'cyan,green,gray' npm:client npm:api npm:queue",
+    "test": "npm run client:test",
```
demiankatz, line 35: We should probably document this one in the README as well (I assume that this will eventually run both api and client tests in parallel, once we have api tests established)

package.json
```diff
@@ -27,10 +28,11 @@
     "backend:wsl": "concurrently npm:api:wsl npm:queue:wsl",
     "build": "concurrently npm:api:build npm:client:build",
     "dev": "concurrently -c 'cyan,green,gray' npm:client npm:api:dev npm:queue:dev",
-    "format": "prettier --check --write api/*.js api/src/** client/src/**",
+    "format": "prettier --check --write api/*.js api/src/** client/src/*.{js,jsx}",
```
demiankatz: This is fine for now, but would it make sense to include ts and tsx as well in case we decide to use Typescript in the React code?

demiankatz (commented):
Thanks for all the progress, this is very nearly polished to perfection... but I have a few more minor questions and suggestions from my latest review.

client/src/Category.test.js
diboy2: oops

client/src/Category.test.js
diboy2: initial value is false if session storage didn't previously exist.

client/src/Category.test.js
diboy2: made change

client/src/Category.test.js
diboy2: made change.

demiankatz: Ahh, I see, because the control starts out closed. But it might be more clear if there's than a corresponding toBeTruthy check lower down, after we've clicked to open it.

client/package.json
diboy2: removed

demiankatz:
Thanks, @diboy2 -- it all looks good to me, and the build is passing, so I'd be fine with merging this. @Geoffsc, do you want to take one last look and then do the honors?

package.json
diboy2: Made chagne.

package.json
diboy2: changed.

client/src/Category.test.js
diboy2: Yeah. It's there in another way. as the toHaveBeenCalled check

PR #75, opened 2021-08-30
Refactor Set 1 of React Components

diboy2:
Still in progress


client/package.json
```diff
@@ -18,6 +18,7 @@
     "start": "react-scripts start",
     "build": "react-scripts build",
     "test": "react-scripts test --coverage --watchAll=false",
+    "testWatch": "react-scripts test --watchAll=true",
```
demiankatz, line 21: Just a reminder to add a corresponding README entry if this is something we're going to keep.

client/src/JobList.test.js
```diff
@@ -0,0 +1,38 @@
...
+            data: ["testJob"],
+        };
+    });
+
+    it("renders", () => {
+        const wrapper = render(<JobList {...props} />);
+        expect(toJson(wrapper)).toMatchSnapshot();
+        expect(mockJob).toHaveBeenCalledWith(
+            expect.objectContaining({
+                category: props.category,
```
demiankatz, line 32: May just be my lack of understanding of how mocking works in Jest, but should we be checking that a proper key property is also passed in here? And where is `children: "testJob"` coming from exactly?

client/src/JobSelector.jsx
```diff
@@ -1,56 +1,57 @@
...
-    }
+    };
+
+    var categories = [];
+    var empty_categories = [];
+    const categoryComponents = Object.keys(data).reduce((categoryComponents, key) => {
+        const category = data[key];
+        const categoryComponent = <Category key={category.category} data={category} />;
+        if(category.jobs.length > 0) {
+            categoryComponents[0] = [...categoryComponents[0], categoryComponent];
```
demiankatz: I wonder if it would be a little less confusing if you used a different name for the `categoryComponents` callback function parameter, and the `categoryComponents` constant that the result of the `reduce` operation is assigned to. The reused name caused me some momentary confusion when reading this code.

demiankatz: Also, would `categoryComponents[0].push(categoryComponent)` be functionally equivalent yet more concise, or am a missing a subtle reason why this approach is preferable?

client/src/JobSelector.jsx
```diff
@@ -1,56 +1,57 @@
...
+        } else {
+            categoryComponents[1] = [...categoryComponents[1], categoryComponent];
+        }
+        return categoryComponents;
+    },[[],[]]).reduce((a,b) => [...a, ...b],[]);
+
+    // for (var i in this.state.data) {
```
demiankatz: I think these comments can be removed now (though they were helpful for reference during review, I confess!)

client/src/JobSelector.jsx
```diff
@@ -1,56 +1,57 @@
...
+    // }
+    return (
+        <div className={this.state.active ? "" : "hidden"} id="jobSelector">
+            {categoryComponents}
+        </div>
+    );
+
 }
 
 JobSelector.propTypes = {
```
demiankatz, line 32: Do these propTypes definitions still mean anything after conversion to functional components? Do they need to be updated or removed? I think @crhallberg added them in the first place, so he may know more about their purpose if it is not already clear.

client/src/LogoutButton.test.js
```diff
@@ -0,0 +1,35 @@
...
+
+    it("renders", () => {
+        const wrapper = shallow(<LogoutButton />);
+        expect(toJson(wrapper)).toMatchSnapshot();
+    });
+
+    it("clears the token", () => {
+        const component = mount(<LogoutButton />);
+
+        expect(sessionStorage.removeItem).not.toHaveBeenCalledWith("open-testCategory", "false");
```
demiankatz: I think the parameters to this assertion are copied and pasted from elsewhere, and need adjusting here.

demiankatz (commented):
Thanks for all the work here! I haven't had time to do a full review yet (and I'm not sure if you're fully ready for one), but I left a few initial comments. I also notice that the build is currently failing (I think due to Prettier issues).

Let me know if/when you're ready for further review, and I'll be happy to devote some more time to this.

client/src/JobList.test.js
diboy2: It's coming from the mockJob call. It mocks the job component and

client/src/LogoutButton.test.js
diboy2: Sorry. this isn't ready. I posted this pr just to view what changes I made.

demiankatz: No worries! Just let me know when it's ready for a review and I'll take a look.

client/src/JobSelector.jsx
diboy2: sorry. I'm just used to doing this. There are cases where this would be better.

client/src/JobSelector.jsx
diboy2: Removed

client/src/JobSelector.jsx
diboy2: The linter was complaining about them, so I kept them.

client/src/JobSelector.test.js
```diff
@@ -0,0 +1,37 @@
...
+    "./Category",
+    () =>
+        function Category(props) {
+            mockCategory(props);
+            return <mock-Category />;
+        }
+);
+
+describe("JobSelector", () => {
+    // let data;
```
diboy2, line 19: There's an api call i intend to test with the categories and jobs.

client/package.json
diboy2: done

client/src/JobList.test.js
demiankatz: So mockJob() returns "testJob" because that's what gets passed in to it? And do you think it is worth checking the key property too, or is that not present for some reason?

client/src/JobSelector.test.js
demiankatz: Do you want to finish this test before merging the PR? If not, would it be better to put in a TODO comment so we don't lose track of it later?

client/src/AjaxHelper.js
```diff
@@ -2,7 +2,7 @@ import $ from "jquery";
 
 class AjaxHelperInstance {
     constructor() {
-        this.url = null;
+        this.url = "http://localhost:9000"; // TODO: Config
```
demiankatz, line 5: If we're initializing this here, should we remove the assignment in index.js for now so we don't have this TODO in two places?

client/src/JobSelector.jsx
demiankatz: I see... but I suspect they are no longer accurate. Something we should probably revisit with @crhallberg later; maybe worth putting a task on the board to review them at the end of the refactoring, especially if we're able to reduce/simplify.

client/src/PdfGenerator.test.js
```diff
@@ -0,0 +1,22 @@
...
+    });
+
+    it("changes the pid input", () => {
+        const wrapper = mount(<PdfGenerator />);
+        wrapper.find("#pdfGeneratePid").simulate("change", {
+            target: {
+                value: "testPid",
+            },
+        });
+        expect(wrapper.find("#pdfGeneratePid").instance().value).toEqual("testPid");
```
demiankatz, line 20: Is this test actually testing any of the behavior of the React component itself? It looks to me like you're setting the value of an input, and then confirming that the value of the input changed... but that doesn't seem to confirm whether any of the related React logic actually did anything. As always, apologies if I'm missing the point here. :-)

client/src/SolrIndexer.test.js
```diff
@@ -0,0 +1,22 @@
...
+
+describe("SolrIndexer", () => {
+    it("renders", () => {
+        const wrapper = shallow(<SolrIndexer />);
+        expect(toJson(wrapper)).toMatchSnapshot();
+    });
+
+    it("changes the pid input", () => {
+        const wrapper = mount(<SolrIndexer />);
+        wrapper.find("#solrIndexPid").simulate("change", {
```
demiankatz, line 15: Same comment here as for PdfGenerator.test.js.

demiankatz (commented):
Thanks again for all the progress -- once again, I like what I see here. Refactoring the routes to a second component is a good idea, and you've made some of my sillier code actually look like proper React. ;-)

I've done some hands-on testing on my test server, and everything still seems to function as expected.

I just left a few minor comments, questions and suggestions...

client/src/AjaxHelper.js
diboy2: Made change

client/src/JobSelector.test.js
diboy2: I'll put it as TODO since it is dependent on the refactor of the AjaxHelper class.

diboy2: Made change.

client/src/PdfGenerator.test.js
diboy2: I send an onChange event in the previous line. After onChange, it should call setPid and update pid state. After update of pid state, on the next render because of state change, pid value is passed to the input value attribute into input.

client/src/SolrIndexer.test.js
diboy2: refer to previous comment.

client/src/PdfGenerator.test.js
demiankatz: Okay, great, that makes sense -- thanks for the clarification!

client/src/JobSelector.jsx
```diff
@@ -1,57 +1,39 @@
...
-        return (
-            <div className={this.state.active ? "" : "hidden"} id="jobSelector">
-                {categories}
-                {empty_categories}
-            </div>
-        );
-    }
-}
+const JobSelector = () => {
+    const [active, setActive] = useState(true);
```
demiankatz: I'm not sure if there's actually a reason to have the active state anymore -- before we used React router, there was logic to switch back and forth between components by hiding and unhiding things... but now that we're using proper routing, none of that is necessary. I think this may be a relic of that time.

Geoffsc: @demiankatz seems worth removing this and testing just to make sure state persists the same way

diboy2: So we don't want to hide the jobs anymore? This active property sets the className of the job selector.

diboy2: Oh. Wait. Now I see. This doesn't control hiding the category componet.s

diboy2: Okay. I'm going to remove it.

demiankatz: Exactly, it just hides the whole component... and since the component is likely empty and invisible until data loads, hiding it until data loads seems redundant.

PR #76, opened 2021-09-04
Refactor logic to build objects.

demiankatz:
This PR consolidates and simplifies object creation, with the intent of making the server-side implementation of #72 significantly simpler.

TODO:
- [x] Test everything

### DIFF
--- a/api/src/jobs/GeneratePdf.ts
+++ b/api/src/jobs/GeneratePdf.ts
@@ -3,6 +3,7 @@ import fs = require("fs");
 import Config from "../models/Config";
 import { FedoraObject } from "../models/FedoraObject";
 import { execSync } from "child_process";
+import FedoraObjectFactory from "../services/FedoraObjectFactory";
 import http = require("needle");
 import PDFDocument = require("pdfkit");
 import QueueJobInterface from "./QueueJobInterface";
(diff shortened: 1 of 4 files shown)


demiankatz:
I've been thinking about how our ingest process works, and in the existing code, there are a lot of manual steps to put together objects which I think are better consolidated into a single function, to increase consistency and simplify calls. This PR attempts to address that problem, and this will serve as a foundation for #72.

The idea is that every object should have three models: the core model (which every object has), the type model (which is either "Collection" or "Document"), and the specific model (Page, PDF, Folder, List, etc., etc.). The old code had us adding these manually one at a time. The new code infers the first two models from the third model, using the configuration as a guide.

We had a lot of methods for adding models, but most of them did very little. A couple of them added sort attributes, but I've just moved that into a switch statement in the single big "initialize object" function.

There is probably still room for improvement here, but I think this is a little tighter and more comprehensible than before.

I'm interested in everyone's thoughts when time permits! Once everyone is happy with this, we can merge it into #72 and wire up the object creation there (along with some validation, since we make some assumptions about integrity during ingest that we cannot safely make when dealing with arbitrary user input).

demiankatz:
A bit more thought on this led me to the conclusion that while I'm refactoring, I should also move the pid generation logic out of the model and into the new factory service -- it reduces the code further and feels like a better fit. See bdfef4d4dfc4e12b6ca844a4d5b7ba3f2b40d799.

PR #77, opened 2021-09-08
Receive Camel Toolbox messages

demiankatz:
This PR is WORK IN PROGRESS on receiving messages from the Camel Toolkit into our API.

TODO
- [x] Handle create/update messages
- [x] Handle delete messages


demiankatz:
I've done a bit more work on this, so there is now an actual mechanism for queueing up index actions (with some associated refactoring). I added some deduplication logic to prevent excessive indexing, though I'm not sure how well this will scale in high-volume situations. Obviously, testing will be required.

For the moment, the endpoint is listening for a PID in the message body. That's not what we actually want to do in reality, so once the Camel Toolbox stuff is ready to go, we'll have to revise this to look at the appropriate headers. At that time, we can also remove the body parser, because I don't think we'll actually need to do anything with the body at all. We'll also need to add some logic to do different things for different types of actions (e.g. deletes vs. creates vs. updates). Right now, the system doesn't have a mechanism for handling deletes, but this should be easy enough to add.

PR #78, opened 2021-09-09
Refactor Set 2 of React Components

diboy2:


client/src/ZoomToggleButton.jsx
```diff
@@ -1,15 +1,9 @@
...
-    render() {
-        return (
-            <button onClick={this.props.paginator.toggleZoom}>
-                {this.props.paginator.state.zoom ? "Turn Zoom Off" : "Turn Zoom On"}
-            </button>
-        );
-    }
-}
+const ZoomToggleButton = ({ paginator }) => {
+    return <button onClick={paginator.toggleZoom}>{paginator.state.zoom ? "Turn Zoom Off" : "Turn Zoom On"}</button>;
```
demiankatz: Extremely trivial point, but we could make the line shorter by factoring "Turn Zoom" out of the two strings and only making Off/On conditional.

demiankatz (commented):
Apart from the problem with thumbnail scrolling we're discussing in Teams, the only comment I had was very minor. I haven't thoroughly reviewed the Thumbnail and PaginatorList components yet since I expect they may change while this bug is fixed.

client/src/ZoomToggleButton.jsx
diboy2: Made change

demiankatz:
Thanks again, @diboy2 -- this all looks good to me (and works as expected). I imagine there is further work that could be done to simplify the relationship between the PaginatorList and Thumbnail components, but this is incremental progress and I assume that will be part of future work on state management.

crhallberg (approved):
The PaginatorControlGroup refactor in particular is really elegant and instructive for me! Thank you!

PR #79, opened 2021-09-21
Refactor JobPaginator Component

diboy2:
Anything with AjaxHelper I'm probably going to refactor and probably split into a client model and shared state model.

### DIFF
--- a/client/src/JobPaginator.jsx
+++ b/client/src/JobPaginator.jsx
@@ -1,326 +1,73 @@
-import React from "react";
+import React, { useEffect } from "react";
 import PropTypes from "prop-types";
 
-import AjaxHelper from "./AjaxHelper";
-import MagicLabeler from "./MagicLabeler";
+import JobPaginatorZoomToggle from "./JobPaginatorZoomToggle";
 import PaginatorControls from "./PaginatorControls";
 import PaginatorList from "./PaginatorList";
-import PaginatorPreview from "./PaginatorPreview";
-import PaginatorZoomy from "./PaginatorZoomy";
+import useJobPaginator from "./useJobPaginator";
 
-class JobPaginator extends React.Component {
-    constructor(props) {
-        super(props);
-        this.magicLabelCache = [];
-        this.state = { active: false, currentPage: 0, zoom: false, order: [] };
-        this.ajax = AjaxHelper.getInstance();
-        // BIND
-        this.autonumberFollowingPages = this.autonumberFollowingPages.bind(this);
-        this.getLabel = this.getLabel.bind(this);
-        this.toggleZoom = this.toggleZoom.bind(this);
-        this.deletePage = this.deletePage.bind(this);
-        this.nextPage = this.nextPage.bind(this);
-        this.prevPage = this.prevPage.bind(this);
-    }
+const JobPaginator = ({ initialCategory, initialJob }) => {
+    const {
+        state: { currentPage, zoom, order, category, job },
+        action: {
+            setLabel,
+            getLabel,
+            getMagicLabel,
+            setPage,
+            nextPage,
+            prevPage,
+            deletePage,
+            save,
+            loadJob,
+            autonumberFollowingPages,
+            getImageUrl,
+            setZoom,
+        },
+    } = useJobPaginator(initialCategory, initialJob);
 
-    getImageUrl(imageNumber, size) {
-        if (typeof this.state.order[imageNumber] === "undefined") {
-            return false;
-        }
-        this.filename = this.state.order[imageNumber].filename;
-        return this.ajax.getImageUrl(this.state.category, this.state.job, this.filename, size);
-    }
+    useEffect(() => {
+        loadJob();
+    }, []);
 
-    getStatusUrl() {
-        return this.ajax.getJobUrl(this.state.category, this.state.job, "/status");
-    }
-
-    getLabel(imageNumber, useMagic) {
-        useMagic = typeof useMagic === "undefined" ? true : useMagic;
-        var label =
-            typeof this.state.order[imageNumber] === "undefined" ? null : this.state.order[imageNumber]["label"];
-        if (useMagic && null === label) {
-            if (typeof this.magicLabelCache[imageNumber] === "undefined") {
-                this.magicLabelCache[imageNumber] = MagicLabeler.getLabel(imageNumber, this.getLabel);
-            }
-            return this.magicLabelCache[imageNumber];
-        }
-        // Always return a string, even if the internal value is null:
-        return label;
-    }
-
-    setLabel(imageNumber, text) {
-        this.magicLabelCache = []; // clear label cache whenever there is a change
-        var newState = this.state;
-        if (text !== null && text.length === 0) {
-            text = null;
-        }
-        if (typeof newState.order[imageNumber] === "undefined") {
-            return;
-        }
-        newState.order[imageNumber]["label"] = text;
-        this.setState(newState);
-        dispatchEvent(new Event("Prep.editted"));
-    }
-
-    autonumberFollowingPages() {
-        var pages = this.state.order.length - (this.state.currentPage + 1);
-        var affected = pages - this.countMagicLabels(this.state.currentPage + 1);
-        if (affected > 0) {
-            var msg = "You will be clearing " + affected + " label(s). Are you sure?";
-            if (!window.confirm(msg)) {
-                return;
-            }
-        }
-        for (var i = this.state.currentPage + 1; i < this.state.order.length; i++) {
-            this.setLabel(i, null);
-        }
-    }
-
-    countMagicLabels(startAt) {
-        var count = 0;
-        for (var i = startAt; i < this.state.order.length; i++) {
-            if (null === this.getLabel(i, false)) {
-                count++;
-            }
-        }
-        return count;
-    }
-
-    deletePage() {
-        if (this.state.order.length < 2) {
-            alert("You cannot delete the last page in a job.");
-            return;
-        }
-        if (!window.confirm("Are you sure you wish to delete the current page?")) {
-            return;
-        }
-        var imageUrl = this.getImageUrl(this.state.currentPage, "*");
-        var parts = imageUrl.split("/");
-        var imageFilename = parts[parts.length - 2];
-        this.ajax.ajax({
-            type: "DELETE",
-            url: imageUrl,
-            success: function () {
-                this.removePages([imageFilename]);
-                if (this.state.currentPage >= this.state.order.length) {
-                    this.setPage(this.state.currentPage - 1);
-                }
-                alert("Page deleted!");
-            }.bind(this),
-            error: function () {
-                alert("Unable to delete!");
-            },
-        });
-    }
-
-    loadJob(category, job) {
-        var promise = new Promise(
-            function (resolve /*, reject*/) {
-                this.ajax.getJSON(this.ajax.getJobUrl(category, job), null, function (data) {
-                    resolve(data);
-                });
-            }.bind(this)
-        );
-        promise
-            .then(
-                function (data) {
-                    data.category = category;
-                    data.job = job;
-                    data.active = false;
-                    data.currentPage = 0;
-                    this.setState(data);
-                    return new Promise(
-                        function (resolve /*, reject*/) {
-                            this.ajax.getJSON(this.getStatusUrl(), null, function (data) {
-                                resolve(data);
-                            });
-                        }.bind(this)
-                    );
-                }.bind(this)
-            )
-            .then(
-                function (status) {
-                    if (status.file_problems.deleted.length > 0 || status.file_problems.added.length > 0) {
-                        var msg = "";
-                        if (status.file_problems.deleted.length > 0) {
-                            msg +=
-                                status.file_problems.deleted.length +
-                                " file(s) have been removed = require(the job since the last edit.\n)";
-                            this.removePages(status.file_problems.deleted);
-                        }
-                        if (status.file_problems.added.length > 0) {
-                            msg +=
-                                status.file_problems.added.length +
-                                " file(s) have been added to the job since the last edit.\n";
-                            this.addPages(status.file_problems.added);
-                        }
-                        alert(msg);
-                    }
-                    var newState = this.state;
-                    newState.active = true;
-                    this.setState(newState);
-                    dispatchEvent(new Event("Prep.loaded"));
-                }.bind(this)
-            );
-    }
-
-    findNewPagePosition(page, list) {
-        for (var i = 0; i < list.length; i++) {
-            if (list[i].filename >= page) {
-                return i;
-            }
-        }
-        return i;
-    }
-
-    addPages(pages) {
-        var newState = this.state;
-        for (var i = 0; i < pages.length; i++) {
-            newState.order.splice(this.findNewPagePosition(pages[i], newState.order), 0, {
-                filename: pages[i],
-                label: null,
-            });
-        }
-        this.setState(newState);
-    }
-
-    removePages(pages) {
-        var newOrder = [];
-        for (var i = 0; i < this.state.order.length; i++) {
-            var include = true;
-            for (var j = 0; j < pages.length; j++) {
-                if (this.state.order[i].filename === pages[j]) {
-                    include = false;
-                    break;
-                }
-            }
-            if (include) {
-                newOrder[newOrder.length] = this.state.order[i];
-            }
-        }
-        var newState = this.state;
-        newState.order = newOrder;
-        this.setState(newState);
-    }
-
-    setPage(p) {
-        if (p >= 0 && p < this.state.order.length) {
-            var newState = this.state;
-            newState.currentPage = p;
-            this.setState(newState);
-        }
-    }
-
-    nextPage() {
-        this.setPage(this.state.currentPage + 1);
-    }
-
-    prevPage() {
-        this.setPage(this.state.currentPage - 1);
-    }
-
-    saveMagicLabels() {
-        for (var i = 0; i < this.state.order.length; i++) {
-            if (null === this.getLabel(i, false)) {
-                this.setLabel(i, this.getLabel(i));
-            }
-        }
-    }
-
-    confirmSavedMagicLabels(count) {
-        const msg = "You will be saving " + count + " unreviewed, auto-generated label(s). Are you sure?";
-        return window.confirm(msg);
-    }
-
-    save(publish) {
-        var count = this.countMagicLabels(0);
-        if (count > 0 && !this.confirmSavedMagicLabels(count)) {
-            return;
-        }
-        this.saveMagicLabels();
-        var promise = new Promise(
-            function (resolve /*, reject*/) {
-                // If the user wants to publish, let's make sure all derivatives are
-                // ready! Otherwise we can resolve with no further actions.
-                if (publish) {
-                    this.ajax.getJSON(this.getStatusUrl(), null, function (data) {
-                        resolve(data);
-                    });
-                } else {
-                    resolve(null);
-                }
-            }.bind(this)
-        );
-        promise.then(
-            function (data) {
-                if (publish) {
-                    var msg;
-                    if (data.derivatives.expected > data.derivatives.processed) {
-                        msg =
-                            "Derivative images have not been generated yet. Please" +
-                            ' go back to the main menu and hit the "build" button' +
-                            " for this job before publishing it.";
-                        alert(msg);
-                        return;
-                    }
-                    msg =
-                        "Are you sure you wish to publish this job? You will not be able" +
-                        " to make any further edits.";
-                    if (!window.confirm(msg)) {
-                        return;
-                    }
-                }
-                this.ajax.ajax({
-                    type: "PUT",
-                    url: this.ajax.getJobUrl(this.state.category, this.state.job, ""),
-                    contentType: "application/json",
-                    data: JSON.stringify({ order: this.state.order, published: publish }),
-                    success: function () {
-                        alert("Success!");
-                        window.location.assign("/paginate"); // TODO: Route better?
-                        dispatchEvent(new Event("Prep.saved"));
-                    },
-                    error: function () {
-                        alert("Unable to save!");
-                    },
-                });
-            }.bind(this)
-        );
-    }
-
-    toggleZoom() {
-        this.newState = this.state;
-        this.newState.zoom = !this.newState.zoom;
-        this.setState(this.newState);
-    }
-
-    componentDidMount() {
-        this.loadJob(this.props.category, this.props.job);
-    }
-
-    render() {
-        var preview = this.state.zoom ? (
-            <PaginatorZoomy img={this.getImageUrl(this.state.currentPage, "large")} />
-        ) : (
-            <PaginatorPreview img={this.getImageUrl(this.state.currentPage, "medium")} />
-        );
-        return (
-            <div className={this.state.active ? "" : "hidden"} id="paginator">
-                <div className="row">
-                    <div className="six col">{preview}</div>
-                    <div className="six col">
-                        <p>
-                            {this.state.category} &gt; {this.state.job}
-                        </p>
-                        <PaginatorControls paginator={this} />
-                        <PaginatorList paginator={this} pageCount={this.state.order.length} />
-                    </div>
+    return (
+        <div id="paginator">
+            <div className="row">
+                <div className="six col">
+                    <JobPaginatorZoomToggle zoom={zoom} getImageUrl={(size) => getImageUrl(order[currentPage], size)} />
+                </div>
+                <div className="six col">
+                    <p>
+                        {category} &gt; {job}
+                    </p>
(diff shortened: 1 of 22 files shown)


client/src/JobPaginator.jsx
```diff
@@ -1,326 +1,64 @@
...
+            deletePage,
+            save,
+            autonumberFollowingPages,
+            getImageUrl,
+            setZoom,
+        },
+    } = useJobPaginator(initialCategory, initialJob);
+
+    return (
+        <div className={active ? "" : "hidden"} id="paginator">
```
demiankatz: I suspect that "active" can be totally eliminated here, since (as with one of the other components you worked on) it was only here to hide the component in the days before we introduced routing to the application.

client/src/JobPaginatorState.js
```diff
@@ -0,0 +1,305 @@
...
+            message.push(added.length + " file(s) have been added to the job since the last edit.\n");
+            setOrder(getAddedPages(order, added));
+        }
+        return message.length ? alert(message.join()) : undefined;
+    };
+
+    const loadJob = async () => {
+        const { order } = await getJob(category, job);
+        setOrder(order);
+        setCategory(category);
```
demiankatz: Is there any point to the setCategory and setJob calls here? It seems to me they're just setting the values to their current values, but maybe I'm missing a subtlety.

client/src/JobPaginator.jsx
demiankatz: (I do see that this is apparently used to hide everything while data is loading, so maybe that's useful... but I'd be interested to see if there's a negative consequence to disabling those calls).

client/src/JobPaginatorState.js
```diff
@@ -0,0 +1,305 @@
...
+        setPage(currentPage - 1);
+    };
+
+    const updatePagesByStatus = (status) => {
+        const {
+            file_problems: { deleted, added },
+        } = status;
+        const message = [];
+        if (deleted.length > 0) {
+            message.push(deleted.length + " file(s) have been removed = require(the job since the last edit.\n)");
```
demiankatz: This should be "have been removed from the job" -- looks like perhaps a search-and-replace or other IDE feature malfunctioned in the past. (I realize that this problem was not introduced by this PR, but we might as well fix it while we see it). Here's the original code from the previous iteration of the project: https://github.com/FalveyLibraryTechnology/VuDLPrep/blob/f59a7529aa4bb6afecd4508ca5284637a23ee3c9/app/assets/javascripts/components/JobPaginator.js.jsx#L120

client/src/JobPaginatorZoom.jsx
```diff
@@ -0,0 +1,22 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+import PaginatorPreview from "./PaginatorPreview";
+import PaginatorZoomy from "./PaginatorZoomy";
+
+const JobPaginatorZoom = ({ zoom, getImageUrl }) => {
```
demiankatz: This might need a better name, since it's only a Zoom when zoom is set to true. Maybe JobPaginatorZoomToggle?

demiankatz: Or JobPaginatorTogglableZoom?

client/src/JobPaginatorZoom.jsx
```diff
@@ -0,0 +1,22 @@
...
+    return (
+        <div className="six col">
```
demiankatz: I also wonder whether it's better to factor this outermost `<div>` back into the Job Paginator itself... since this is part of the paginator's layout, it seems like it might be cleaner to keep the layout columns together in the same component so it's easier to adjust them, rather than embedding them in this component (which reduces its ability to be used in a different context).

demiankatz:
@diboy2, I left a few comments, though I still haven't done a fully comprehensive review. I also tried to run the code currently in the branch, but it does not appear to be in a working state quite yet -- I think there are some problems related to the label code. No labels display on any of the thumbnails, and if I try to set a label, I get a fatal error:

<pre>
TypeError: order[imageNumber] is undefined
setLabel
src/JobPaginatorState.js:141
</pre>

I realize this may still be work in progress, so I won't put more time into testing until you give me the go-ahead. In the meantime, I'll leave a couple more comments related to labels which might be helpful.

client/src/JobPaginatorState.js
```diff
@@ -0,0 +1,305 @@
...
+        }
+        order[imageNumber]["label"] = text !== null && text.length === 0 ? null : text;
+        setOrder(order);
+        dispatchEvent(new Event("Prep.editted"));
+    };
+
+    const getMagicLabel = (imageNumber) => {
+        const label = getLabel(order, imageNumber);
+
+        if (!label) {
```
demiankatz: Is this logic backwards? It seems like it will cause this to always return null (or at least return null under inappropriate circumstances).

I appreciate your efforts to make the getLabel() method less confusing and more explicit, but I think some subtlety has gotten lost along the way. There are really three scenarios that need to be addressed:

1.) We want to know whether or not the user has set a label explicitly on an image.
2.) We want to get the auto-generated "magic label" for a set position.
3.) We want to get the user-generated label if available, or the "magic label" otherwise.

I think in the current code, cases 2 and 3 are getting mixed up and it's causing problems. Maybe we want three functions: getUserLabel, getMagicLabel, and getBestLabel... that might make this all more clear.

client/src/Thumbnail.jsx
```diff
@@ -1,22 +1,18 @@
...
     }, [selected]);
 
     return (
-        <div onClick={selectPage} className={"thumbnail" + (selected ? " selected" : "")} ref={ref}>
-            <img alt="" src={paginator.getImageUrl(number, "thumb")} />
+        <div onClick={() => setPage(number)} className={"thumbnail" + (selected ? " selected" : "")} ref={ref}>
+            <img alt="" src={getImageUrl(number, "thumb")} />
             <div className="number">{number + 1}</div>
-            <div className={"label" + (null === paginator.getLabel(number, false) ? " magic" : "")}>{label}</div>
+            <div className={"label" + (!getMagicLabel(number) ? " magic" : "")}>{getMagicLabel(number)}</div>
```
demiankatz: Here's an example of a possible problem with label use. I think what we might really want here is something like:

`<div className={"label" + (null === getUserLabel(number) ? " magic" : "")}>{getBestLabel(number)}</div>`

In other words, display the most appropriate label for the thumbnail, but if it's an auto-generated magic label, apply an extra class so we can style it differently.

client/src/JobPaginatorState.js
demiankatz: I also realized that the "\n" is in the wrong place in this string (something @crhallberg fixed in #71, but which hasn't been merged anywhere else).

client/src/useJobPaginator.js
```diff
@@ -0,0 +1,191 @@
...
+    const getMagicLabel = (imageNumber) => {
+        const label = getLabel(order, imageNumber);
+
+        if (label) {
+            return label;
+        }
+        if (typeof magicLabelCache[imageNumber] === "undefined") {
+            magicLabelCache[imageNumber] = MagicLabeler.getLabel(imageNumber, getMagicLabel);
+        }
+        setMagicLabelCache(magicLabelCache);
```
demiankatz: Is there a reason this was moved out of the if statement above? I would think that we only need to set this value when it changes, but maybe there's a subtlety I'm not aware of.

client/src/JobPaginatorState.js
diboy2: had the same thought. this was in the old code. I wasn't sure why it was there, so I kept it there in case I broke something.

client/src/JobPaginatorState.js
diboy2: Yes. I realized that after applying the unit tests. I reserved the logic

client/src/JobPaginatorZoom.jsx
diboy2: I think I should have kept it in the job paginator after reviewing the code.

client/src/useJobPaginator.js
diboy2: I was trying to fix tsom things. I'm putting it back.

client/src/JobPaginator.jsx
diboy2: let me test it now.

client/src/JobPaginatorState.js
demiankatz: I believe that in the old code, the assignment was necessary because "category" was coming in as a function parameter and getting assigned to the state. In the new code, you're initializing the state via the "initialCategory" property, so there's no longer a need to reassign the value later.

client/src/JobPaginator.jsx
```diff
@@ -1,326 +1,69 @@
...
+            <div className="row">
+                <JobPaginatorZoom zoom={zoom} getImageUrl={(size) => getImageUrl(order[currentPage], size)} />
+                <div className="six col">
+                    <p>
+                        {category} &gt; {job}
+                    </p>
+                    <PaginatorControls
```
diboy2, line 42: This is still overly wordy for me. I'm planning to move some states and functions to a shared state, since they are redundantly used in the child components anyway.

client/src/JobPaginatorZoom.jsx
diboy2: actually, I'll move the column

client/src/JobPaginatorZoom.jsx
diboy2: Making change.

client/src/JobPaginatorZoom.jsx
diboy2: Made change.

client/src/JobPaginator.jsx
diboy2: made change. I'm not sure if it's okay or not.

client/src/JobPaginatorState.js
diboy2: Okay. I removed it

client/src/JobPaginatorState.js
diboy2: I made the change.

client/src/JobPaginator.jsx
demiankatz: Yes, that makes sense, though if you want to keep this as-is for now and work on that in a separate iteration, that would be fine with me. Might be good to get the current working state merged before making further revisions.

client/src/JobPaginator.jsx
demiankatz: There is a brief flash as the data loads, where the control switches from an empty state to a populated state... but I think that's okay for now. In the longer-term, I think a better solution is to make this control behave more correctly when the order is empty. There are some jobs which do not have any pages (e.g. audio jobs, PDF-only jobs), and it would be nice to have a more sensible display in those cases. I think it may make sense to pass the order.length into the PaginatorControls and hide all of the page-specific controls when there are no pages. Similarly, it may make sense for the PaginatorList to display a message like "No pages found." in place of the thumbnail list when pageCount is empty. This is not urgent, though -- maybe the best thing for now is to create a card for it and deal with it later.

client/src/JobPaginator.jsx
```diff
@@ -1,326 +1,71 @@
...
+            deletePage,
+            save,
+            useEffect,
+            autonumberFollowingPages,
+            getImageUrl,
+            setZoom,
+        },
+    } = useJobPaginator(initialCategory, initialJob);
+
+    useEffect();
```
demiankatz: What is the purpose of an empty call to useEffect()? I haven't seen that before (but I'm fairly new to this).

client/src/JobPaginator.test.js
```diff
@@ -0,0 +1,52 @@
...
+    });
+
+    it("renders", () => {
+        const wrapper = shallow(<JobPaginator {...props} />);
+        expect(wrapper.contains(props.initialCategory)).toBeTruthy();
+        expect(wrapper.contains(props.initialJob)).toBeTruthy();
+        expect(toJson(wrapper)).toMatchSnapshot();
+    });
+
+    it("should loadJob from useEffect", async () => {
```
demiankatz, line 45: Isn't this happening as a result of useJobPaginator rather than useEffect?

client/src/JobPaginatorState.js
```diff
@@ -0,0 +1,129 @@
...
+const getStatus = (category, job) => {
+    const ajax = AjaxHelper.getInstance();
+    return new Promise((resolve) =>
+        ajax.getJSON(ajax.getJobUrl(category, job, "/status"), null, (data) => {
+            resolve(data);
+        })
+    );
+};
+
+const deleteImage = (imageUrl, success, error) => {
```
demiankatz, line 32: I just tested and found that the "delete image" feature doesn't work on this branch for some reason. I think that imageUrl is not getting constructed correctly, because when I look at my network console, I see DELETE requests going to URLs like: `http://localhost:9000/api/ingest/dime/job2/undefined/6`. Obviously, if you want to test this, you'll want to make a copy of your example job to test with so that you don't lose your original test data.

client/src/JobPaginatorState.js
```diff
@@ -0,0 +1,129 @@
...
+        if (!getLabel(order, startIndex)) {
+            count++;
+        }
+    }
+    return count;
+};
+
+const validateSave = (order) => {
+    const count = countMagicLabels(order, 0);
+    return count > 0 && !confirmSavedMagicLabels(count);
```
demiankatz, line 110: The confirmSavedMagicLabels logic is inverted here; I have to hit "cancel" in order to save, and if I hit OK, the save action is canceled. (You can test this by trying to save a job with auto-generated labels).

demiankatz: On a possibly-related note, apart from having inverted logic, the "Save" button does seem to work correctly. However, the "Save and Publish" button is more broken. I can't seem to find a way to publish a job at all, and during my attempts, I get this error:

<pre>
Warning: Cannot update a component (`JobPaginator`) while rendering a different component (`Thumbnail`). To locate the bad setState() call inside `Thumbnail`, follow the stack trace as described in https://fb.me/setstate-in-render
    in Thumbnail (at PaginatorList.jsx:18)
    in div (at PaginatorList.jsx:35)
    in PaginatorList (at JobPaginator.jsx:55)
    in div (at JobPaginator.jsx:36)
    in div (at JobPaginator.jsx:32)
    in div (at JobPaginator.jsx:31)
    in JobPaginator (at Routes.jsx:12)
    in JobPaginatorHook (at Routes.jsx:25)
    in Route (at Routes.jsx:24)
    in Switch (at Routes.jsx:17)
    in Routes (at VuDLPrep.jsx:15)
    in Router (created by BrowserRouter)
    in BrowserRouter (at VuDLPrep.jsx:14)
    in div (at VuDLPrep.jsx:10)
    in VuDLPrep (at src/index.js:14)
    in StrictMode (at src/index.js:13)
</pre>

This is followed by some weird, malformed API requests.

demiankatz (changes requested):
@diboy2, I spent some more time testing and reviewing this PR today. It seems that the label-related functionality is now in very good shape, but some of the other features have issues. I haven't finished comprehensively reviewing all code, but since I found some significant bugs, I thought I would stop here and let you work on fixing these before I finish up, since I expect there will be some further changes now. Please let me know if you need help reproducing any of the issues I've reported!

client/src/JobPaginator.jsx
diboy2: useEffect replaces componentDidMount, and componentDidUpdate, I repurposed useEffect in useJobPaginator.

diboy2: I think I'll just move loadJob into useEffect, to make this more explicit.

client/src/JobPaginatorState.js
diboy2: Made change.

client/src/JobPaginator.jsx
diboy2: Moved it.

client/src/JobPaginatorState.js
diboy2: I modified and made changes.

client/src/JobPaginator.jsx
diboy2: https://teams.microsoft.com/l/entity/com.microsoft.teamspace.tab.planner/tt.c_19:54c0a65e7f5f491aa6f90021330eb74e@thread.tacv2_p_LH9rYXUCYke47Igm88R2rGQAGFab_h_1612357088580?tenantId=765a8de5-cf94-44f0-9caf-ae5bf8cfa366&webUrl=https%3A%2F%2Ftasks.teams.microsoft.com%2Fteamsui%2FpersonalApp%2Falltasklists&context=%7B%22subEntityId%22%3A%22%2Fboard%2Ftask%2F-fOapF7EMEyz0VaK771xXmQAI2KZ%22%2C%22channelId%22%3A%2219%3A54c0a65e7f5f491aa6f90021330eb74e%40thread.tacv2%22%7D

diboy2: https://teams.microsoft.com/l/entity/com.microsoft.teamspace.tab.planner/tt.c_19:54c0a65e7f5f491aa6f90021330eb74e@thread.tacv2_p_LH9rYXUCYke47Igm88R2rGQAGFab_h_1612357088580?tenantId=765a8de5-cf94-44f0-9caf-ae5bf8cfa366&webUrl=https%3A%2F%2Ftasks.teams.microsoft.com%2Fteamsui%2FpersonalApp%2Falltasklists&context=%7B%22subEntityId%22%3A%22%2Fboard%2Ftask%2F-fOapF7EMEyz0VaK771xXmQAI2KZ%22%2C%22channelId%22%3A%2219%3A54c0a65e7f5f491aa6f90021330eb74e%40thread.tacv2%22%7D

diboy2: Created these two tasks.

client/src/JobPaginatorState.js
```diff
@@ -0,0 +1,127 @@
...
+    return pages.reduce((order, page) => {
+        order.splice(findNewPagePosition(page, order), 0, {
+            filename: page,
+            label: null,
+        });
+        return order;
+    }, order);
+};
+
+const getNoneRemovedPages = (order, deletedPages) =>
```
demiankatz: Is this a typo in the function name? Should it be getNonRemovedPages? Or would getRemainingPages be more clear?

client/src/useJobPaginator.js
```diff
@@ -0,0 +1,186 @@
...
+                },
+                () => {
+                    alert("Unable to delete!");
+                }
+            );
+        }
+    };
+
+    const save = async (published) => {
+        if (validateSave(order)) {
```
demiankatz: I think validateSave is another function in need of a better name. It's a bit confusing that it returns false when we should proceed and true when we should not -- I would expect the opposite from something with "validate" in the name. I would either reverse the return value of the function, or rename it to something more specific like `userAbortedSave` or `userMustReviewMoreLabels`.

client/src/JobPaginatorState.js
```diff
@@ -0,0 +1,127 @@
...
+    deleteImage,
+    deletePageValidation,
+    getAddedPages,
+    getNoneRemovedPages,
+    getStatus,
+    getJob,
+    getLabel,
+    putJob,
+    confirmSavedMagicLabels,
+    findNewPagePosition,
```
demiankatz: findNewPagePosition seems to be an internal support method; do we need to export it?

client/src/PaginatorControls.jsx
```diff
@@ -32,13 +32,11 @@ class PaginatorControls extends React.Component {
             useMagic = true;
         }
         var label = this.labelInput.value;
-        return label.length === 0 && useMagic
-            ? this.props.paginator.getLabel(this.props.paginator.state.currentPage)
-            : label;
+        return label.length === 0 && useMagic ? this.props.getMagicLabel(this.props.currentPage) : label;
```
demiankatz, line 35: I see that this component is still class-based; I assume you plan to work on converting this in a future iteration.

client/src/Thumbnail.jsx
```diff
@@ -1,22 +1,18 @@
...
     }, [selected]);
 
     return (
-        <div onClick={selectPage} className={"thumbnail" + (selected ? " selected" : "")} ref={ref}>
-            <img alt="" src={paginator.getImageUrl(number, "thumb")} />
+        <div onClick={() => setPage(number)} className={"thumbnail" + (selected ? " selected" : "")} ref={ref}>
+            <img alt="" src={getImageUrl(number, "thumb")} />
             <div className="number">{number + 1}</div>
-            <div className={"label" + (null === paginator.getLabel(number, false) ? " magic" : "")}>{label}</div>
+            <div className={"label" + (!getLabel(number) ? " magic" : "")}>{getMagicLabel(number)}</div>
```
demiankatz: Is it safe to change the `null ===` condition to a simple `!` here? Is there a risk that falsy values will get misinterpreted as magic labels here and get the `magic` class inappropriately applied? (I tried to make this happen and couldn't, so it's probably fine -- but just wanted to raise the possibility in case there's some edge case you can think of). If this actually is a problem, there are a few other places that might need adjustment as well.

client/src/Thumbnail.test.js
```diff
@@ -29,20 +28,21 @@ describe("Thumbnail", () => {
...
+        expect(props.getImageUrl).toHaveBeenCalledWith(props.number, "thumb");
         expect(wrapper.find(".thumbnail.selected").exists()).toBeTruthy();
         expect(wrapper.text().includes("13")).toBeTruthy();
         expect(wrapper.find(".label.magic").exists()).toBeTruthy();
     });
 
     it("renders correctly when not selected or no label", () => {
         props.selected = false;
-        props.paginator.getLabel.mockReturnValue("testLabel");
+        props.getLabel.mockReturnValue(true);
```
demiankatz: It might be more realistic to also return "testLabel" here -- generally getLabel is either going to return null or the same value as getMagicLabel.

client/src/useJobPaginator.js
```diff
@@ -0,0 +1,186 @@
...
+        if (deleted.length > 0) {
+            message.push(deleted.length + " file(s) have been removed from the job since the last edit.\n");
+            setOrder(getNoneRemovedPages(order, deleted));
+        }
+
+        if (added.length > 0) {
+            message.push(added.length + " file(s) have been added to the job since the last edit.\n");
+            setOrder(getAddedPages(order, added));
+        }
+        return message.length ? alert(message.join()) : undefined;
```
demiankatz: Does it really make sense to have a `return` here at all? While it's a tidy one-liner as written, it's also a bit unusual. I wonder if it would be more straightforward to just have a standard `if (message.length) { alert(message.join()); }` here.

client/src/useJobPaginator.test.js
```diff
@@ -0,0 +1,299 @@
...
+            });
+
+            act(() => {
+                result.current.action.setLabel(0, "test2");
+            });
+
+            expect(result.current.state.order[0].label).toEqual("test2");
+        });
+
+        it("does not change the label", () => {
```
demiankatz: This is testing that set doesn't do anything if you specify an invalid index, correct? Maybe the description should incorporate that, like "does not change the label if index out of range".

demiankatz (changes requested):
Thanks again for all the work on this, @diboy2 -- I think it's very nearly done. Just a few very minor comments about naming, test descriptions, and possible (but I think unlikely) edge case issues.

client/src/JobPaginatorState.js
diboy2: I thought nonRemovedPages was more appropriate?

client/src/JobPaginatorState.js
diboy2: Oh, yes. I'll make that change.

client/src/PaginatorControls.jsx
diboy2: Yes. JobPaginator is already large enough of a task. It's not in the scope.

client/src/JobPaginatorState.js
diboy2: since deletedPages was one of the parameters

client/src/Thumbnail.jsx
diboy2: I'll change it back to null

client/src/Thumbnail.test.js
diboy2: this might have been from a previous test.

client/src/useJobPaginator.js
diboy2: I don't know why I did it that way. I think it's because I watned it to return messages initially.

client/src/useJobPaginator.js
diboy2: if(userMustReviewMoreLabels)

client/src/JobPaginatorState.js
diboy2: Changing spelling.

client/src/useJobPaginator.test.js
diboy2: Made change.

demiankatz (approved):

PR #82, opened 2021-09-23
Add npm script to update snapshots.

demiankatz:
TODO:
- [x] Make the script actually work right

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,7 @@ After a few moments, a new tab should automatically open in your browser pointin
 | client:build | build React code for production |
 | client:lint | lint only the api code |
 | client:setup | install npm dependencies in client/ |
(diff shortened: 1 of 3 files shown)


demiankatz:
@diboy2, I need to regenerate snapshots to make the tests pass in #72. I thought it would be useful to have an npm script to do this (and so it could be more easily documented). However, the code as written here isn't working, and I'm not sure what I'm missing. Can you point me in the right direction?

demiankatz:
Thanks for your help on our call, @diboy2 -- if you like this as-is, feel free to hit merge when time permits. If you think it would be better for the script name to include "test" I'm happy to adjust. Just being lazy for now and keeping things mostly the way I set them up originally.

PR #83, opened 2021-09-23
Refactor metadata extraction to a separate class.

demiankatz:
As I think about the next steps on #72, it occurs to me that it would be useful to have access to metadata extraction capabilities independent of the hierarchy collector. It is also better separation of concerns to split hierarchy collection and metadata extraction into separate classes. This PR takes care of the split, adds some missing function comments, and removes an unused class property.


demiankatz:
The new MetadataExtractor class would also be an excellent candidate for a unit test; we should start building out the api-side framework for that soon.

PR #85, opened 2021-09-29
Add queue visibility using Arena.

demiankatz:
TODO
- [x] Require login to access Arena

### DIFF
--- a/api/app.js
+++ b/api/app.js
@@ -9,12 +9,13 @@ const session = require("express-session");
 const path = require("path");
 const logger = require("morgan");
 
-const authRouter = require("./dist/routes/auth").router;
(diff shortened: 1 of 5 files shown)


demiankatz:
This was a pretty simple integration, but we need to figure out how to integrate it with Passport authentication. I tried using the authenticate() method defined in routes/auth.ts, but that didn't redirect to the right place. I think we may need to do some refactoring of the authentication logic to get this to all fit together more neatly.

demiankatz:
I was able to get the login requirement implemented, though it required some bug-fixing and refactoring. I'd be interested to know if @crhallberg approves of my changes.

crhallberg (approved):

PR #87, opened 2021-10-06
Add shared state. Add fetch context. Refactor Job

diboy2:
This is not complete, but I wanted to show some progress.


client/src/Job.jsx
```diff
@@ -1,212 +1,31 @@
...
+    } = useJob({ category, children });
+
+    useEffect(() => {
+        updateStatus();
+    }, []);
+
+    return (
+        <li>
+            <JobClickable category={category} clickable={clickable} clickWarning={clickWarning}>
+                children
```
demiankatz: Should this be `{children}`?

client/src/useJobHelper.test.js
```diff
@@ -0,0 +1,19 @@
...
+
+import { getAgeString } from "./useJobHelper";
+
+describe("useJobHelper", () => {
+    test.each([
+        [1, "1 minute old"],
+        [2, "2 minutes old"],
+        [61, "1 hour old"],
+        [121, "2 hours old"],
+        [1440, "1 day old"],
```
demiankatz, line 11: Should we have a test for 2 days old? Seems like the only thing missing from the pattern here.

client/src/__snapshots__/PdfGenerator.test.js.snap
```diff
@@ -1,29 +1,17 @@
...
+    Object {
+      "dispatch": [Function],
+      "state": Object {
+        "previousParams": null,
+        "token": null,
+      },
+    }
+  }
+>
+  <PdfGenerator />
```
demiankatz: Is it intended that the PdfGenerator snapshot no longer actually includes the rendered content of the PDF Generator? Has the introduction of the ContextProvider had unanticipated side effects on snapshot testing?

demiankatz (commented):
Thanks, @diboy2, this is looking very good to me at first glance, though I confess I haven't spent a lot of time looking closely yet. Just passing along a few early observations. Let me know when you would like a more thorough review and/or some hands-on testing, and I'll be happy to provide more feedback.

client/src/Job.jsx
diboy2: Fixed that.

client/package.json
```diff
@@ -23,6 +22,7 @@
     "testWatch": "react-scripts test --coverage --watchAll=true",
     "eject": "react-scripts eject"
   },
+  "proxy": "http://localhost:5000",
```
demiankatz: Out of curiosity, what is this used for?

client/src/SolrIndexer.jsx
```diff
@@ -1,25 +1,17 @@
...
-            dataType: "text",
-            error: function (result, status) {
-                setResults(status);
-            },
-            success: function (result) {
-                setResults(result);
-            },
-        });
-    }
+    const doApiCall = async (method) => {
```
demiankatz, line 11: Something seems to be wrong with this tool. When I try to preview an index operation, I get:

<pre>
Error: Objects are not valid as a React child (found: [object Promise]). If you meant to render a collection of children, use an array instead.
    in pre (at SolrIndexer.jsx:25)
    in div (at SolrIndexer.jsx:17)
    in SolrIndexer (at Routes.jsx:31)
    in Route (at Routes.jsx:18)
    in Switch (at Routes.jsx:17)
    in Routes (at VuDLPrep.jsx:16)
    in Router (created by BrowserRouter)
    in BrowserRouter (at VuDLPrep.jsx:15)
    in FetchContextProvider (at VuDLPrep.jsx:11)
    in VuDLPrep (at src/index.js:14)
    in StrictMode (at src/index.js:13)
</pre>

client/src/context.js
```diff
@@ -0,0 +1,125 @@
...
+const fetchReducer = (state, { type, payload }) => {
+    switch (type) {
+        case "UPDATE_TOKEN":
+            sessionStorage.setItem("token", payload);
+            return {
+                ...state,
+                token: payload,
+            };
+        default:
+            console.error("action type does not exist");
```
demiankatz: Would it be helpful to actually output the value of type as part of the error message in this case, so if it ever happens we know which action was requested?

client/src/context.js
```diff
@@ -0,0 +1,125 @@
...
+        mode: "cors",
+        ...params,
+        ...(params.headers ? params.headers : {}),
+    });
+
+    const refreshToken = async () => {
+        try {
+            const response = await fetch(`${authApiUrl}/token/mint`, fetchParams({}));
+            if (response.ok) {
+                console.log("it was okay");
```
demiankatz: Guessing we don't want to keep this console.log indefinitely. :-)

demiankatz:
@diboy2, thanks for the progress; I've left a few scattered comments (but haven't done a thorough review yet). For now, I'm just playing with it to see if I can make it break. In addition to the notes above, there seems to be a problem with the login follow-up system. Here is the expected behavior, which I am seeing on the dev branch:

1.) Go to main menu, not logged in
2.) Click on "Job Paginator"
3.) Get prompted to log in
4.) Log in
5.) Get redirected to job paginator screen

In this branch, at step 5, I get redirected back to the main menu instead of forward to the job paginator after I log in. When the application redirects to the login page, it is passing the correct referer parameter to the /api/auth/login route... so I'm not sure why things aren't behaving as expected following the login. But since this works in dev and not here, there must be some difference that is causing the change in behavior. Please let me know if you need help troubleshooting!

demiankatz:
@diboy2, one other important regression I've noticed so far: the job selector no longer seems to respect published status. When I hit the "save and publish" button, it edits the job.json in my job directory to set the "published" property to true, but the job selection list lets me continue to edit the job and does not present the "ingest now" button.

client/package.json
diboy2: I don't know why that's there.

demiankatz: Strange! If I'm understanding the documentation correctly (which I might not be), this is used to tell NPM to use an HTTP proxy for outgoing requests. I don't think we would want to do this by default, and I'm not sure how it would have gotten introduced to your configuration.

client/src/SolrIndexer.jsx
diboy2: Made change.

diboy2: forgot to call await

client/src/useJobHelper.test.js
diboy2: Made change.

client/package.json
diboy2: I just removed it and it's working now.

client/src/context.js
diboy2: Made change

client/src/context.js
diboy2: Made change.

client/src/__snapshots__/PdfGenerator.test.js.snap
diboy2: Made change.

client/src/useJob.test.js
```diff
@@ -0,0 +1,49 @@
...
+    it("sets the click warning if less than 10 minutes", async () => {
+        statusResponse.minutes_since_upload = 5;
+
+        const { result } = await renderHook(() => useJob(props), { wrapper: FetchContextProvider });
+
+        await act(async () => {
+            await result.current.action.updateStatus();
+        });
+
+        expect(result.current.state.statusText).toEqual(["loading..."]);
```
demiankatz: This is the same assertion as the previous test; is there supposed to be an assertion related to the click warning?

client/src/JobPaginator.test.js
```diff
@@ -14,39 +14,42 @@ jest.mock("./JobPaginatorZoomToggle", () => () => "JobPaginatorZoomToggle");
...
-            file_problems: {
-                deleted: [],
-                added: [],
-            },
-        });
+        global.fetch = jest.fn(() =>
+            Promise.resolve({
+                json: () =>
+                    Promise.resolve({
+                        order: "testOrder",
```
demiankatz: It might be more realistic, especially in light of #89, for the order fixture to be an array rather than a string.

client/src/JobPaginator.test.js
```diff
@@ -14,39 +14,42 @@ jest.mock("./JobPaginatorZoomToggle", () => () => "JobPaginatorZoomToggle");
...
         await act(async () => {
-            await mount(<JobPaginator {...props} />);
+            await mount(
+                <FetchContextProvider>
+                    <JobPaginator {...props} />
+                </FetchContextProvider>
+            );
         });
-        expect(JobPaginatorState.getJob).toHaveBeenCalled();
-        expect(JobPaginatorState.getStatus).toHaveBeenCalled();
```
demiankatz, line 50: Does this test do anything if you've removed the assertions?

client/src/JobPaginatorState.test.js
```diff
@@ -1,20 +1,10 @@
 import { beforeEach, describe, expect, it, jest } from "@jest/globals";
 import * as JobPaginatorState from "./JobPaginatorState";
-import AjaxHelper from "./AjaxHelper";
-jest.mock("./AjaxHelper");
 
 describe("JobPaginatorState", () => {
-    let ajax;
```
demiankatz, line 7: Since `findNewPagePosition` has effectively been made public, should a test for it be added here?

client/src/LogoutButton.jsx
```diff
@@ -1,14 +1,13 @@
...
+import { logoutUrl } from "./routes";
 
 const LogoutButton = () => {
-    const ajax = AjaxHelper.getInstance();
     const clearToken = () => {
         sessionStorage.removeItem("token");
     };
     return (
-        <a href={ajax.logoutUrl} className="button" onClick={clearToken}>
+        <a href={logoutUrl} className="button" onClick={clearToken}>
```
demiankatz, line 10: Out of scope for this PR, but for the to-do list: we shouldn't display the log out button when we don't have a token; maybe we actually need a log in button that can trigger the log in process manually.

client/src/PdfGenerator.jsx
```diff
@@ -1,25 +1,17 @@
...
-            error: function (result, status) {
-                setResults(status);
-            },
-            success: function (result) {
-                setResults(result);
-            },
-        });
+    const doApiCall = async (method) => {
+        const url = baseUrl + "/messenger/pdfgenerator/" + encodeURIComponent(pid);
+        setResults(await getText(url, { method }));
```
demiankatz: Do we need better error handling here? It looks to me like getText will return an error object if an exception is thrown; do we need to do type checking? Or is there a better approach?

client/src/SolrIndexer.jsx
```diff
@@ -1,25 +1,17 @@
...
+        const url = baseUrl + "/messenger/solrindex/" + encodeURIComponent(pid);
+        setResults(await getText(url, { method }));
```
demiankatz: Same error handling question as above.

client/src/context.js
```diff
@@ -0,0 +1,128 @@
+import React, { createContext, useContext, useReducer } from "react";
```
demiankatz, line 1: General comment: since there are some complex workflows in this file, would it be helpful to add some comments? (Not a requirement -- just thinking there may be some orders of operations and caveats that are worth capturing while they're fresh in mind, so this is easier to understand when revisited in the future).

client/src/context.test.js
```diff
@@ -0,0 +1,58 @@
+import { beforeEach, describe, expect, it, jest } from "@jest/globals";
+import { renderHook, act } from "@testing-library/react-hooks";
+import { FetchContextProvider, useFetchContext } from "./context";
+
+describe("useFetchContext", () => {
```
demiankatz, line 5: Should tests be added for getText and getJSON? Should these tests cover the error cases that I expressed concerns about above?

client/src/routes.js
```diff
@@ -0,0 +1,40 @@
+const baseUrl = "http://localhost:9000";
+const apiUrl = `${baseUrl}/api`;
+const authApiUrl = `${apiUrl}/auth`;
+const ingestApiUrl = `${apiUrl}/ingest`;
+const loginUrl = `${authApiUrl}/login`;
+const logoutUrl = `${authApiUrl}/logout`;
+const testLoginAsChris = `${authApiUrl}/user/confirm/V1StGXR8_Z5jdHi6B-myT`;
```
demiankatz: Since `testLoginAsChris` is a temporary hack for testing purposes, would it make more sense to definite locally where it is used rather than putting it here?

client/src/useJob.js
```diff
@@ -0,0 +1,133 @@
...
+        const percentDone = 100 * (derivatives.processed / derivatives.expected);
+        setClickable(true);
+        return ["derivatives: " + percentDone.toFixed(2) + "% built"];
+    };
+
+    const getJobStatusText = ({ derivatives, documents, audio, ingesting, published }) => {
+        if (derivatives.expected === 0 && documents === 0 && audio === 0) {
+            return ["empty job"];
+        }
+        const pageCount = parseInt(derivatives.expected / 3);
```
demiankatz: The magic number "3" here refers to the number of derivative types generated per page. Maybe we should define this as a constant (`derivativeTypeCount`, maybe) to make the code a little more readable.

client/src/useJob.test.js
```diff
@@ -0,0 +1,55 @@
...
+    });
+
+    it("sets the click warning if less than 10 minutes", async () => {
```
demiankatz, line 42: Does this description really reflect what is happening in the test?

demiankatz (commented):
Thanks, @diboy2 -- I've read this much more thoroughly, and this is great work. You've done an excellent job of untangling the knots in this code. Most of my comments are minor and refer to tests. I think the biggest remaining issue may be to make sure that errors are handled gracefully in the case of unexpected HTTP behavior... but maybe that's not as big an issue as I think.

In any case, I'm going to do a little more end-to-end testing in my test environment to make sure the whole process still works from derivative generation through ingestion. I'll report back if I encounter any problems.

demiankatz:
@diboy2, I found a bug: the action button doesn't clear when triggered. Here are steps to reproduce:

1.) Create a new job in your holding area (for example, I just created a "job4" folder inside my dime folder in the holding area and copied 001.tif and 002.tif into it from a previous job).

2.) Click "create derivatives"

3.) Now derivatives start building, but the button does not disappear as it should while the build process is taking place.

demiankatz:
The ingest process has two problems, one of which is clearly related to the issue above. If I publish my job after derivatives are generated, I get the "Ingest now" button. When I click it, it does not disappear as it should (just like "Create derivatives"). Additionally, while the job is ingesting, I should get a real-time status line showing what is happening, but this does not automatically appear -- I have to refresh the page to get the updates to start appearing.

Let me know if you need help setting up enough of the test environment to see this in action.

client/src/JobPaginator.test.js
diboy2: Made change

client/src/JobPaginator.test.js
diboy2: Yes. I added the test.

client/src/JobPaginatorState.test.js
diboy2: Made change.

client/src/LogoutButton.jsx
diboy2: Sure. Made task. https://teams.microsoft.com/l/entity/com.microsoft.teamspace.tab.planner/tt.c_19:54c0a65e7f5f491aa6f90021330eb74e@thread.tacv2_p_LH9rYXUCYke47Igm88R2rGQAGFab_h_1612357088580?tenantId=765a8de5-cf94-44f0-9caf-ae5bf8cfa366&webUrl=https%3A%2F%2Ftasks.teams.microsoft.com%2Fteamsui%2FpersonalApp%2Falltasklists&context=%7B%22subEntityId%22%3A%22%2Fboard%2Ftask%2FSQuRlSQ2ZkygdjkfMadb0WQAG-en%22%2C%22channelId%22%3A%2219%3A54c0a65e7f5f491aa6f90021330eb74e%40thread.tacv2%22%7D

client/src/PdfGenerator.jsx
diboy2: Reverted it back.

client/src/SolrIndexer.jsx
diboy2: Reverted it back.

client/src/context.test.js
diboy2: Added.

client/src/routes.js
diboy2: I'm not even sure if it's being used anywhere. I don't see it being referenced inthe client code.

demiankatz: Well, if you remove it and the "just log in as chris" button continues to work, then I think we can safely say it's not needed. :-)

demiankatz:
@diboy2, thanks for the progress on this. Things are working better now with the action buttons. However, there was a problem where clearing out the text fields on every status update was causing a very distracting blinking behavior. I think we only need to clear the statuses if something goes wrong, so I moved those lines -- see commit bee86cf.

demiankatz:
@diboy2, when I click into a job from the job selector, I'm getting this in my console:

<pre>
Warning: Cannot update a component (`JobPaginator`) while rendering a different component (`Thumbnail`). To locate the bad setState() call inside `Thumbnail`, follow the stack trace as described in https://fb.me/setstate-in-render
    in Thumbnail (at PaginatorList.jsx:18)
    in div (at PaginatorList.jsx:35)
    in PaginatorList (at JobPaginator.jsx:60)
    in div (at JobPaginator.jsx:41)
    in div (at JobPaginator.jsx:34)
    in div (at JobPaginator.jsx:33)
    in JobPaginator (at Routes.jsx:12)
    in JobPaginatorHook (at Routes.jsx:25)
    in Route (at Routes.jsx:21)
    in Switch (at Routes.jsx:17)
    in Routes (at VuDLPrep.jsx:16)
    in Router (created by BrowserRouter)
    in BrowserRouter (at VuDLPrep.jsx:15)
    in FetchContextProvider (at VuDLPrep.jsx:11)
    in VuDLPrep (at src/index.js:14)
    in StrictMode (at src/index.js:13) index.js:1

​</pre>

crhallberg (approved):

client/src/routes.js
diboy2: I just removed it, and i'm still able to login as chris.

client/src/useJob.js
diboy2: made change

client/src/useJob.test.js
diboy2: Made change.

client/src/context.js
diboy2: Made change.

demiankatz (approved):

PR #88, opened 2021-10-14
Create tool for retrieving object children.

demiankatz:
This PR is designed to make the hierarchy tree navigable in the editor.

TODO
- [x] Implement initial API for retrieving children
- [x] Refine/expand/polish parameters/fields in API endpoint
- [x] Implement route/React component for displaying/navigating children
- [x] Write tests


demiankatz:
I've made good progress on implementing the React part of this -- it's _almost_ working. However, when you navigate two levels deep into the tree, the contents of the tree do not refresh. Apparently I'm not understanding how `useEffect` and React router interact with each other. I'll do some more research soon to see if I can sort this out.

demiankatz:
Fixed my problem: apparently the issue was that my ObjectEditor wasn't being assigned a unique key in the ObjectEditorHook, which was preventing routing from causing a re-render. I'm not sure why this is a problem here but not for the JobPaginator, but maybe it's because you can link from ObjectEditor to ObjectEditor, but not from JobPaginator to JobPaginator.

I'm not seeing exactly the hierarchy I'm expecting, but this may be either an indexing bug or a flaw in my query logic. I'll investigate that next.

demiankatz:
Okay, I think I have all the indexing issues sorted out, and this is now behaving as expected. Next step: write tests.

demiankatz:
I made progress on tests today, though the ChildList test isn't generating correct snapshots for some reason -- this will require a bit more investigation when I have more time. (I've confirmed that the simulated fetch call is working correctly, but for some reason the retrieved data isn't appearing in the snapshot -- I suspect it is some kind of timing issue).

demiankatz:
Tests fully working now, thanks to help from @diboy2 (the key issue is that you need to call `.update()` on your wrapper in the test if you want snapshots to reflect DOM changes after the initial load when using `mount`). Requesting a review now -- as far as I can tell, this is nearly ready to merge.

api/src/routes/edit.ts
```diff
@@ -2,9 +2,32 @@ import express = require("express");
...
+import Solr from "../services/Solr";
 
 router.get("/models", requireToken, function (req, res) {
     res.json({ CollectionModels: Config.getInstance().collectionModels, DataModels: Config.getInstance().dataModels });
 });
 
+async function getChildren(req, res) {
+    const query =
+        (req.params.pid ?? "").length > 0
+            ? 'fedora_parent_id_str_mv:"' + req.params.pid.replace('"', "") + '"'
```
diboy2: you can `"${variable}"`. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

api/src/services/Solr.ts
```diff
@@ -32,6 +32,18 @@ class Solr {
...
     }
 
+    public async query(
+        core: string,
+        solrQuery: string,
+        queryParams: Record<string, string> = {}
+    ): Promise<NeedleResponse> {
+        let extras = "";
+        for (const i in queryParams) {
+            extras += "&" + encodeURIComponent(i) + "=" + encodeURIComponent(queryParams[i]);
```
diboy2: `Object.values(queryParams).reduce((extras, value) => { return extras + "";}),"");`

diboy2:
Not done reviewing. I might forgot.

api/src/routes/edit.ts
demiankatz: Good point -- improved in 598aea2.

api/src/services/Solr.ts
demiankatz: I don't think a reduce operation can work here because we need to account for both keys and values... however, for a more functional twist on the logic, I tried this instead, which I think is an improvement: d1b5bd5

demiankatz:
Thanks, @diboy2, I have responded to both of your suggestions. Please let me know if you have further feedback, or if you think this is now ready to merge.

api/src/services/SolrIndexer.ts
```diff
@@ -134,6 +141,7 @@ class SolrIndexer {
                 }
             }
         }
+        fields.fedora_parent_id_str_mv = fedoraParents.map((parent) => parent.pid);
```
diboy2, line 144: what does this key mean? I usually avoid abbreviations since the meaning could be ambiguous to the someone maintaining the code. mv could mean move or multi-value, or market value.

demiankatz: This is a Solr field name. The abbreviations are required, because they are built into the Solr schema.

client/src/Editor/ChildList.jsx
```diff
@@ -0,0 +1,40 @@
...
+const ChildList = ({ pid = null }) => {
+    const {
+        action: { fetchJSON },
+    } = useFetchContext();
+    const [children, setChildren] = useState([]);
+
+    useEffect(() => {
+        async function loadData() {
+            let data = [];
+            const url = apiUrl + "/edit/object/children" + (pid === null ? "" : "/" + pid);
```
diboy2, line 16: u can fetch something without a pid?

demiankatz: The service provides the children of the provided pid. If you do not provide a pid, it returns the top-level objects that have no parents. I considered giving this an entirely different URL route (and am still open to the possibility), but I couldn't think of a great name for that, and using a more unified routing strategy kept things simple.

api/src/services/SolrIndexer.ts
diboy2: ok

diboy2 (approved):
👍

PR #89, opened 2021-10-19
Hide page controls when there are no pages.

demiankatz:
TODO:
- [x] Update tests


demiankatz:
@diboy2, this may not be the most elegant solution, and I'm sure it needs some test coverage added, but this is the functionality I'm looking for in terms of hiding page controls when there are no pages available. Feel free to pick it up and finish it if you need something to do; otherwise, I can work on it some more when time permits.

demiankatz:
Also note that I had to update the snapshot to make the tests pass, and it's now testing the "no contents" case. We probably at least want a snapshot test that loads order data and triggers the "normal" controls to render.

demiankatz:
@diboy2, I've made more progress on this -- I've resolved conflicts with the latest dev branch and added testing for the changes to the PaginatorList component.

I haven't bothered with tests for PaginatorControls, since that class still needs to be revised as a functional component, and I assume tests will be implemented at that time.

I am running into one problem I'm not sure how to solve, however: I can't figure out how to get the JobPaginator test to render the JobPaginatorZoomToggle. It always seems to act as though `order.count` is 0, which suggests to me that perhaps the mock fetch isn't working exactly as expected. I did try adding a snapshot test to "should loadJob from useEffect" to see if that had different results than the "renders" test (I expected it to), but it did not. Any ideas?

client/src/JobPaginator.jsx
```diff
@@ -29,15 +29,17 @@ const JobPaginator = ({ initialCategory, initialJob }) => {
         loadJob();
     }, []);
 
+    const preview =
```
diboy2: You could move this property to JobPaginatorZoomToggle.

client/src/PaginatorControls.jsx
```diff
@@ -143,6 +143,26 @@ class PaginatorControls extends React.Component {
                     <i className="fa fa-fw fa-trash"></i> Delete Current Page
                 </button>
             </div>
+        ) : (
+            // No pages, only show save buttons:
+            <div>
+                <button
```
diboy2, line 149: You can just also do `<> </>`

diboy2 (approved):

client/src/PaginatorControls.jsx
demiankatz: Good point -- fixed in 6d84bc7.

client/src/JobPaginator.jsx
demiankatz: I refactored in 34ea41b -- this does feel cleaner.

demiankatz:
Thanks for the review, @diboy2 -- if you're happy with my adjustments, I think this is ready to merge (and then we can close out some related tasks on the job board).

diboy2 (approved):

demiankatz:
Thanks, @diboy2 -- merging now!

PR #91, opened 2021-10-27
Add api unit tests, started env configs

diboy2:


api/.env.review
```diff
@@ -0,0 +1,8 @@
+# URL of React client:
+CLIENT_URL=http://localhost:3000
+
+# Credentials for connecting to a Fedora backend:
+FEDORA_USERNAME=fedoraAdmin
+FEDORA_PASSWORD=fedoraAdmin
+FEDORA_PID_NAMESPACE=vudl
+BASE_URL=http://localhost:8080/rest
```
demiankatz: I would rename this to FEDORA_BASE_URL. It was already bugging me that the .ini file didn't have a more specific configuration name, so moving it here is a perfect opportunity to fix that.

api/src/models/Config.ts
```diff
@@ -20,19 +20,19 @@ class Config {
     }
 
     get clientUrl(): string {
-        return this.ini["client_url"];
+        return process.env.CLIENT_URL;
```
demiankatz: Since we are no longer pulling these values from the .ini file, we should remove them from vudl.ini.dist. We should also update the README to indicate the need for .env and suggest best practices.

Should we also null coalesce these values to the defaults from the .env example file (i.e. `process.env.CLIENT_URL ?? "http://foo"`), so things work if the user skips that configuration step? Those defaults should work in most default dev environments, so I think it could save a little time.

package.json
```diff
@@ -34,7 +36,7 @@
     "lint": "eslint api/*.js api/src/ client/src/ --ext .ts,.js,.jsx --max-warnings 0",
     "setup": "concurrently npm:api:setup npm:client:setup",
     "start": "concurrently -c 'cyan,green,gray' npm:client npm:api npm:queue",
-    "test": "npm run client:test",
+    "test": " npm run client:test",
```
demiankatz: Did you mean to add api:test here as well? Seems like only a blank space has been added for some reason.

api/src/jobs/Index.test.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    describe("run", () => {
+        let job: Job;
+        let indexer: SolrIndexer;
+        let needleResponse: NeedleResponse;
+        beforeEach(() => {
+            needleResponse = {
+                statusCode: 200,
+            } as NeedleResponse;
+            indexer = new SolrIndexer(null, null, null);
```
demiankatz: Would it be better to create a mock SolrIndexer object rather than constructing a "real" one with empty dependencies?

api/src/routes/index.test.ts
```diff
@@ -0,0 +1,14 @@
+import * as request from "supertest";
+import { StatusCodes } from "http-status-codes";
+import app = require("../../app");
+
+describe("index", () => {
+    it("can get the favicon index", async () => {
+        await request(app)
+            .get("/favicon.ico")
+            .expect((response: request.Response) => {
+                expect(response.text).toEqual("go away");
```
demiankatz: Might be better to test for response code 404 here rather than this message text, since Chris was being facetious here and it might get changed. ;-)

api/src/services/DateSanitizer.test.ts
```diff
@@ -0,0 +1,22 @@
+import DateSanitizer from "./DateSanitizer";
+
+describe("DateSanitizer", () => {
+    test.each([
```
demiankatz, line 4: Is there a reason this exists outside of an `it` block? Would it make sense to put this inside `it("sanitizes dates")`, or is there a mechanism for adding descriptions to each of the cases being iterated over? I'd be happy to provide descriptions of individual test cases if that would be useful (though I'm also happy to be lazy if you don't feel there's a need).

client/package.json
```diff
@@ -15,11 +15,11 @@
     "roman-numerals": "^0.3.2"
   },
   "scripts": {
-    "start": "react-scripts start",
+    "start": "REACT_APP_STAGE=dev react-scripts start",
```
demiankatz: I'm not really familiar with best practices here, but does it make sense to hard-code the start task to dev mode? Would production deployment use a totally different mechanism? (Obviously we'll need to sort out those details at some point, and since we're currently focused on dev, I don't object to this for the moment... just wondering if we should consider doing this in a more future-proof way).

demiankatz (changes requested):
Excellent work, once again, @diboy2 -- I think this sets a strong foundation for us to build upon, and we can talk about splitting up test coverage work at this afternoon's meeting. I just had a few questions and comments; see below.

api/package.json
```diff
@@ -3,8 +3,12 @@
   "version": "0.0.0",
   "private": true,
   "scripts": {
+    "api:dev": "nodemon --watch 'api/vudl.ini' --watch 'api/**/*.js' --watch 'api/**/*.ejs' --ignore api/node_modules/ src/app.ts",
```
demiankatz: Should this be `start:dev` (or should `start` be renamed to `api`) to keep a parallel structure between dev and prod tasks? Also, does the current `start` task actually work/do anything? I suspect it may be a relic of boilerplate setup code.

api/worker-start.ts
```diff
@@ -0,0 +1,4 @@
+const JobQueue = require("./dist/services/JobQueue").default; // eslint-disable-line @typescript-eslint/no-var-requires
```
demiankatz: Looks like you've added worker-start.ts but not deleted worker-start.js. Is this intentional?

api/src/app.ts
```diff
@@ -0,0 +1,162 @@
+/* eslint @typescript-eslint/no-var-requires: "off"
```
demiankatz: Is this annotation still needed?

api/src/app.ts
```diff
@@ -0,0 +1,162 @@
...
+   ------
+   Only applies to Typescript files */
+
```
demiankatz: Looks like you've added app.ts but not deleted app.js. Is this intentional?

demiankatz (commented):
Thanks for the ongoing progress here -- just a few questions and comments based on the latest changes.

api/src/services/DateSanitizer.test.ts
diboy2: , You can add another element on the array.

api/src/app.ts
diboy2: I'm going to try to remove it.

api/src/server.ts
```diff
@@ -0,0 +1,85 @@
+/**
+ * Module dependencies.
+ */
+const debug = require("debug")("vudlprepjs:server"); // eslint-disable-line
```
demiankatz: Any idea why eslint needs to be disabled here? Is there a reason we can't change const to import to make it happy (if that's what it's complaining about)?

client/src/config.js
```diff
@@ -0,0 +1,3 @@
+export default {
+    BASE_URL: "http://localhost:9000",
```
demiankatz: Maybe BACKEND_BASE_URL or API_SERVER_BASE_URL? It would probably be helpful to be able to distinguish between the front end and back end URLs, in case there's ever a reason to configure both.

api/src/services/DateSanitizer.test.ts
demiankatz: So something like:

<pre>
        ["n.d.", null, "Unknown date, no space."],
        ["n. d.", null, "Unknown date, with space"],
        ["February 29, 2008", "2008-02-29T00:00:00Z", "Legal leap year date, human readable M D, Y"],
        ["February 29, 2009", "2009-02-01T00:00:00Z", "Illegal leap year date, human readable M D, Y"],
        ["01 June 1467", "1467-06-01T00:00:00Z", "Date far in the past, human readable D M Y"],
        ["1872", "1872-01-01T00:00:00Z", "Just a year"],
        ["[1872]", "1872-01-01T00:00:00Z", "Just a year, in brackets"],
        ["1872&1873", "1872-01-01T00:00:00Z", "Two years joined with ampersand"],
        ["1888-01", "1888-01-01T00:00:00Z", "Year-month, numeric format"],
        ["1888-28", "1888-01-01T00:00:00Z", "Year-illegal month, numeric format"],
        ["1888-05-12", "1888-05-12T00:00:00Z", "Year-month-day, numeric format"],
        ["1888-05-99", "1888-05-01T00:00:00Z", "Year-month-illegal day, numeric format"],
        ["1888--05--12", "1888-05-12T00:00:00Z", "Year--month--day, with double dashes"],
        ["693-01-07", "0693-01-01T00:00:00Z", "Year-month-day, with three-digit year"],
</pre>

api/src/worker-start.ts
```diff
@@ -0,0 +1,4 @@
+const JobQueue = require("./services/JobQueue").default; // eslint-disable-line @typescript-eslint/no-var-requires
```
demiankatz: Does this line need to be cleaned up to make eslint happy, or is it really necessary to do it this way?

demiankatz (changes requested):
Thanks for the progress here! I've resolved some outdated comments and added a few new ones. I left a number of my previous comments open, either because they're still relevant here, or because they might be useful for reference when creating your separate dotenv PR.

Speaking of dotenv, if we're not ready to introduce it yet, should we remove it from api/package.json in this PR and move that to a different one? Or is it still being used here somewhere that I'm overlooking?

api/src/routes/index.test.ts
diboy2: Made change.

api/.env.review
diboy2: I'm moving these changes to another branch.

api/src/models/Config.ts
diboy2: Moved to another branch.

api/src/services/DateSanitizer.test.ts
diboy2: Made change.

api/src/server.ts
diboy2: Made Change

client/src/config.js
diboy2: Made change.

api/src/worker-start.ts
diboy2: Made change

api/src/app.ts
```diff
@@ -0,0 +1,36 @@
+/* eslint @typescript-eslint/no-var-requires: "off"
```
demiankatz: Can changing `const` to `import` in the require lines below eliminate the need for this annotation?

api/src/routes/index.test.ts
```diff
@@ -0,0 +1,16 @@
+/* eslint @typescript-eslint/no-var-requires: "off"
```
demiankatz: I don't think this annotation is needed here. Am I missing something?

api/src/server.ts
```diff
@@ -0,0 +1,133 @@
+/* eslint @typescript-eslint/no-var-requires: "off"
```
demiankatz: I think this is another annotation that is no longer needed.

demiankatz (changes requested):
@diboy2, this is looking great, though I haven't done hands-on end-to-end testing yet. Let me know when you're ready for that.

I've left a few more comments, but they're all essentially the same thing: I think we can further reduce the eslint exception annotations here.

I also still think we may want to rename some of the npm scripts for consistency, but that doesn't necessarily have to be addressed here and now.

api/src/app.ts
diboy2: made change

api/src/routes/index.test.ts
diboy2: made change

api/src/server.ts
diboy2: made change

api/package.json
diboy2: Made change to start:api, and start:queue

api/src/jobs/Index.test.ts
diboy2: Made change.

api/package.json
```diff
@@ -3,8 +3,12 @@
   "version": "0.0.0",
   "private": true,
   "scripts": {
-    "start": "node ./bin/www",
-    "queue": "node worker-start.js"
+    "start:api": "nodemon --watch 'api/vudl.ini' --watch 'api/**/*.js' --watch 'api/**/*.ejs' --ignore api/node_modules/ src/server.ts",
+    "start:queue": "nodemon --watch 'api/vudl.ini' --watch 'api/**/*.js' api/worker-start.ts",
+    "start:server": "ts-node src/server.ts",
+    "start:worker": "ts-node src/worker-start.ts",
```
demiankatz, line 9: I'd still change this a little more -- isn't start:api the dev version of start:server, and start:queue the dev version of start:worker? If I'm understanding correctly, should we rename start:api --> start:serverDev and start;queue --> start:workerDev?

demiankatz (changes requested):
Thanks, @diboy2 -- I think this is nearly done, though I still have some questions about the NPM script naming, and I see that there is at least one test failing in Continuous Integration (in addition to the warning that's been present in the test suite for a little while, about mock component naming conventions).

api/package.json
diboy2: One of them watches for chagnes in the code I think.

demiankatz: True, so maybe renaming them with "Watch" instead of "Dev" would be more meaningful... start:serverWatch and start:workerWatch?

diboy2: Okay. so I'll rename it to that, but I do know the original behavior of dev was to watch api and queue also.

demiankatz: The top-level dev task is concurrently running `npm:client npm:api:dev npm:queue:dev`. I think as long as we don't change the behavior of that one, giving the deeper tasks more meaningful names should be fine. It might be worth asking @crhallberg for input, though, since he did the original NPM setup and might have a better sense than I do of how things are intended to be broken down.

crhallberg (approved):

demiankatz (approved):

PR #92, opened 2021-10-28
Breadcrumb component

Geoffsc:
TODO
- [x] Figure out styling for breadcrumb lists
- Figure out how to prevent breadcrumb area from becoming unreasonably large [deferred to a future iteration; not sure how essential this is until we have real-world cases with lots of breadcrumb trails]
- [x] Convert from .jsx to .tsx after merging #99.
- [x] Add types


api/src/routes/edit.ts
```diff
@@ -1,10 +1,19 @@
...
+const router = express.Router();
+
+const pidSanitizer = sanitizeParameters({ pid: /^[a-zA-Z]+:[0-9]+/ }, /^$/);
 
 router.get("/models", requireToken, function (req, res) {
     res.json({ CollectionModels: Config.getInstance().collectionModels, DataModels: Config.getInstance().dataModels });
 });
 
+router.get("/breadcrumbs/:pid", pidSanitizer, requireToken, async function (req, res) {
+    const fedoraData = await HierarchyCollector.getInstance().getHierarchy(req.params.pid, false);
```
demiankatz: I think putting this in a try..catch block that returns an error message with a 400 status might fix the problem you were having where unexpected input blows up the entire Express application. Might make some of your investigating and troubleshooting more convenient, since you won't have to keep restarting the service every time something goes wrong.

Geoffsc: Yea agree, will update this

Geoffsc: Just pushed an update for this

api/src/routes/edit.ts
```diff
@@ -1,10 +1,25 @@
...
 router.get("/models", requireToken, function (req, res) {
     res.json({ CollectionModels: Config.getInstance().collectionModels, DataModels: Config.getInstance().dataModels });
 });
 
+router.get("/breadcrumbs/:pid", pidSanitizer, requireToken, async function (req, res) {
+    try {
+        const fedoraData = await HierarchyCollector.getInstance().getHierarchy(req.params.pid, false);
+        res.json(fedoraData.getBreadcrumbTrail());
+    } catch (e) {
+        console.error("Error retrieving breadcrumbs: " + e);
```
demiankatz, line 102: I think you should make some kind of response here as well; I'd follow the console.error with something like `res.status(500).send("Unexpected error")`.

(Ideally we should try to figure out what went wrong and give an appropriate code based on the type of the error... i.e. 400 for bad input, 404 if the requested PID doesn't exist, etc., etc., only resorting to 500 if we have no idea why things failed... but for now, 500 is a starting point).

Geoffsc: Yes good point, updated!

demiankatz:
@Geoffsc, some recently merged PRs caused conflicts here. I believe that I have resolved them, but I haven't taken the time to test anything. Can you give it a shot when time permits?

Also note that the Solr indexing code has changed a bit, so you'll want to rebuild your index. I think I gave you the tools for that, but I might be misremembering. If I haven't already, remind me at our next one-on-one and I'll show you how.

demiankatz:
I realized that we had existing code in our VuFind instance for displaying breadcrumb trails, courtesy of @crhallberg, so rather than trying to re-solve that problem, I just ported that logic here with some minor modifications. This may be a bit overcomplicated -- there's a lot of data transformation going on -- but it does seem to have the desired effect. Now the next thing is to figure out how to present these lists in a sensible way.

demiankatz:
I've added some TODOs to reflect remaining work.

api/src/models/FedoraData.ts
```diff
@@ -58,6 +64,16 @@ class FedoraData {
         return results;
     }
 
+    getBreadcrumbTrail(): TreeNode {
```
demiankatz: Maybe getParents would be a better name for this.

api/src/routes/edit.ts
```diff
@@ -84,4 +85,15 @@ async function getChildren(req, res) {
 edit.get("/object/children", requireToken, getChildren);
 edit.get("/object/children/:pid", requireToken, pidSanitizer, getChildren);
 
+edit.get("/breadcrumbs/:pid", pidSanitizer, requireToken, async function (req, res) {
```
demiankatz: Maybe `/object/parents/:pid` would be a better route for this.

demiankatz (commented):
It occurs to me that the data retrieval piece of this is more about parents than breadcrumbs, so maybe we should change some names.

client/components/edit/Breadcrumbs.tsx
```diff
@@ -0,0 +1,147 @@
...
+                queue.push(parent);
+                if (typeof childLookups[parent.pid] === "undefined") {
+                    childLookups[parent.pid] = [current.pid];
+                } else {
+                    childLookups[parent.pid].push(current.pid);
+                }
+            });
+        }
+        return {
+            topNodes: [...new Set(topNodes)], // deduplicate
```
diboy2, line 60: Reduplicate??

client/components/edit/Breadcrumbs.tsx
```diff
@@ -0,0 +1,147 @@
...
+                console.error("Problem fetching breadcrumb data from " + url);
+            }
+            setTreeData(data);
+        }
+        loadData();
+    }, []);
+
+    function generateBreadcrumbTrails(treeData: TreeData, pid: string) {
+        // BFS from top (root id) to target pid
+        const queue: Array<BreadcrumbTrail> = [];
```
diboy2: U can shorten using treeData.map.

client/components/edit/Breadcrumbs.tsx
demiankatz: The intention of this is to remove duplicates from the array by converting it into a set and back again... so "de-duplicate" should be correct. Or am I misunderstanding your question?

client/components/edit/Breadcrumbs.tsx
demiankatz: Good point! I'll fix that in the near future...

demiankatz: Fixed! Thanks again for the suggestion. Let me know if you have any other feedback before we merge this.

diboy2 (approved):

PR #93, opened 2021-11-05
Improve Arena integration.

demiankatz:
This PR improves the integration with Arena in two ways:

1.) It adds a main menu item to expose access to the queue manager

2.) It fixes post-login redirecting to go to the correct place. Since we capture the post-login URL BEFORE redirecting to the login screen, we actually want to store the CURRENT URL, not the PREVIOUS URL, or else we redirect back farther than we're supposed to. I suspect that this bug was never discovered because the code wasn't being used for anything (most of the redirection logic currently takes place client-side in React logic). At least, I couldn't find a situation where my change here made things worse rather than better.


crhallberg (approved):

demiankatz:
Thanks, @crhallberg; I'm going to go ahead and merge this.

PR #95, opened 2021-11-11
Improve error tolerance; prevent over-caching.

demiankatz:


demiankatz:
@diboy2, @Geoffsc -- this PR contains a couple of indexing fixes (including the caching-related one) which were necessary to get the index to update correctly following creation of objects through #72. However, since these fixes are not really directly related to #72, I split them out to a separate pull request.

As @diboy2 suggested on our call, it would probably be a good idea to revisit our datastream caching logic (and determine whether it is even needed at all). However, until we have time to get that done, I would like to merge this so that things will function correctly in the meantime. I will put in a TODO comment as a reminder to find a better solution.

Any objections to merging this as-is?

PR #96, opened 2021-11-15
Auto-generate master metadata

diboy2:


demiankatz:
Thanks, @diboy2, this looks good to me, though I'll review more closely once you've had time to merge the latest code into it.

One possible future optimization: it is possible to get modification dates/times from the Fedora API for datastreams. It might be useful to add logic to the metadata job that fetches the modification dates of the MASTER and MASTER-MD streams. If MASTER-MD is already newer than MASTER, that is probably a sign that we do not need to generate it again. With this in place, we could restore the previous code that added MASTER-MD as part of initial ingest, which would save network traffic by avoiding the need to download data that has just been uploaded, while still allowing subsequent changes to objects to be picked up and to cause appropriate metadata regeneration. I don't think we need to implement that here unless you wish to, but maybe worth capturing somewhere as a potential enhancement.

api/src/routes/messenger.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    let config;
+    let body;
+    let pid;
+    beforeAll(() => {
+        app.use("/messenger", messenger);
+        config = {
+            restBaseUrl: "www.test.com",
+            allowedOrigins: ["http://localhost:3000", "http://localhost:9000"],
+        };
+        pid = "123";
```
demiankatz: Might be worth changing this to `vudl:123` to be more realistic (all pids have an alphanumeric part, a colon, and a numeric part).

api/src/routes/messenger.ts
```diff
@@ -58,9 +58,13 @@ messenger.post("/camel", bodyParser.json(), async function (req, res) {
         action = "Update";
     }
 
+    console.log(`Camel action: ${action} datastream: ${datastream}`);
```
demiankatz: Was this for debugging purposes, or do you think it's worth keeping? If you want to keep it, I wonder if we're better off including the pid as well, or perhaps just displaying the raw request body at the top of the function.

api/src/services/QueueManager.ts
```diff
@@ -35,20 +35,33 @@ class QueueManager {
...
-        let lastPidAction = null;
-        for (let i = 0; i < jobs.length; i++) {
-            if (jobs[i].name === "index" && jobs[i].data.pid === pid) {
-                lastPidAction = jobs[i].data.action;
-                break;
-            }
-        }
-        if (action === lastPidAction) {
+        const queueJob = { pid, action };
+        if (this.isAlreadyAwaitingAction(jobs, "metadata", queueJob)) {
```
demiankatz: Should this be "index" rather than "metadata"?

api/src/services/QueueManager.ts
```diff
@@ -35,20 +35,33 @@ class QueueManager {
...
             console.log("Skipping queue; " + pid + " is already awaiting " + action + ".");
         } else {
             await q.add("index", { pid, action });
         }
         q.close();
     }
+
+    isAlreadyAwaitingAction(jobs: Array<Job>, name: string, { pid, action }: { pid: string; action: string }): boolean {
+        const matchingJob = jobs.find((job) => {
+            return job.name === name && job.data.pid === pid;
```
demiankatz, line 49: I think this may be an oversimplification -- isn't this going to find the first match in the queue? If so, I think what we really want is to find the last match in the queue. In other words, we might have a queue like:

index, delete, index, delete, index, delete, index, delete

If the next action to be queued is index, we want to add it to the end, since after we have deleted a record, we need to bring it back if it is revived. But if we pull "index" off the front of the queue and then drop the job because it matches, the record will incorrectly stay deleted.

Apologies if I'm misunderstanding the code and this is already accounted for.

demiankatz: ...in any case, I guess this makes for a more sophisticated test case that's worth adding!

demiankatz (changes requested):
Thanks, @diboy2 -- overall this looks good, but I think I may have found a bug or two.

api/src/services/QueueManager.ts
diboy2: oh, yes is it is.

api/src/services/QueueManager.ts
diboy2: I misunderstood that check.

diboy2: Okay. the original code looks at the first instance. Maybe we should discuss this?

demiankatz: Ha, you're right, I think that may be a bug on my part, or else I'm misremembering the order of the queue. If it's newest-to-oldest, then it's fine... but if it's oldest-to-newest, then I think it's wrong.

demiankatz:
@diboy2, I tried this out today in my own test environment. I think everything seems correct in terms of the queuing and infrastructure, but I think there is a problem with the actual metadata generation. Specifically, I think the binary data may be getting corrupted as part of the process of getting loaded into a string and written to disk from there. I think we may need to figure out how to stream the data into the file instead of using the intermediate variable loading.

I tested this with a PDF file, and the FITS output when generated through the asynchronous job is significantly different from the FITS output we get if we generate it up-front.

Note in the XML below how the asynchronous job version seems to have almost doubled in size for some reason, and also is no longer recognized as a valid PDF.

Here's the up-front Jhove output:

<pre>
  &lt;fileinfo&gt;
    &lt;size toolname="Jhove" toolversion="1.20.1"&gt;1783416&lt;/size&gt;
    &lt;creatingApplicationName toolname="Exiftool" toolversion="11.54"&gt;Adobe Acrobat Pro DC (32-bit) 21 Paper Capture Plug-in/Adobe Acrobat Pro DC (32-bit) 21.1.20155&lt;/creatingApplicationName&gt;
    &lt;lastmodified toolname="Exiftool" toolversion="11.54" status="CONFLICT"&gt;2021:06:08 23:07:45-04:00&lt;/lastmodified&gt;
    &lt;lastmodified toolname="Tika" toolversion="1.21" status="CONFLICT"&gt;2021-06-09T03:07:45Z&lt;/lastmodified&gt;
    &lt;created toolname="Exiftool" toolversion="11.54" status="SINGLE_RESULT"&gt;2021:06:08 23:03:09-04:00&lt;/created&gt;
    &lt;filepath toolname="OIS File Information" toolversion="1.0" status="SINGLE_RESULT"&gt;/code/holdingarea/pdf/job2/Dollar.pdf&lt;/filepath&gt;
    &lt;filename toolname="OIS File Information" toolversion="1.0" status="SINGLE_RESULT"&gt;Dollar.pdf&lt;/filename&gt;
    &lt;md5checksum toolname="OIS File Information" toolversion="1.0" status="SINGLE_RESULT"&gt;76cbe880500b54334d6f7c75f842f382&lt;/md5checksum&gt;
    &lt;fslastmodified toolname="OIS File Information" toolversion="1.0" status="SINGLE_RESULT"&gt;1623208065941&lt;/fslastmodified&gt;
  &lt;/fileinfo&gt;
  &lt;filestatus&gt;
    &lt;well-formed toolname="Jhove" toolversion="1.20.1" status="SINGLE_RESULT"&gt;true&lt;/well-formed&gt;
    &lt;valid toolname="Jhove" toolversion="1.20.1" status="SINGLE_RESULT"&gt;true&lt;/valid&gt;
  &lt;/filestatus&gt;
</pre>

And here's the version from the asynchronous job:

<pre>
  &lt;fileinfo&gt;
    &lt;size toolname="Jhove" toolversion="1.20.1"&gt;3209070&lt;/size&gt;
    &lt;filepath toolname="OIS File Information" toolversion="1.0" status="SINGLE_RESULT"&gt;/tmp/tmp-5977-7RCIVIs1fm0p&lt;/filepath&gt;
    &lt;filename toolname="OIS File Information" toolversion="1.0" status="SINGLE_RESULT"&gt;tmp-5977-7RCIVIs1fm0p&lt;/filename&gt;
    &lt;md5checksum toolname="OIS File Information" toolversion="1.0" status="SINGLE_RESULT"&gt;94a4f0decea1a330e0e41f92a45c1e9b&lt;/md5checksum&gt;
    &lt;fslastmodified toolname="OIS File Information" toolversion="1.0" status="SINGLE_RESULT"&gt;1637244935826&lt;/fslastmodified&gt;
  &lt;/fileinfo&gt;
  &lt;filestatus&gt;
    &lt;well-formed toolname="Jhove" toolversion="1.20.1" status="SINGLE_RESULT"&gt;false&lt;/well-formed&gt;
    &lt;valid toolname="Jhove" toolversion="1.20.1" status="SINGLE_RESULT"&gt;false&lt;/valid&gt;
    &lt;message toolname="Jhove" toolversion="1.20.1" status="SINGLE_RESULT"&gt;Invalid cross-reference table severity=error offset=1782533&lt;/message&gt;
    &lt;message toolname="Jhove" toolversion="1.20.1" status="SINGLE_RESULT"&gt;No document catalog dictionary severity=error offset=0&lt;/message&gt;
  &lt;/filestatus&gt;
<pre>

api/src/routes/messenger.test.ts
diboy2: Made change.

api/src/routes/messenger.ts
diboy2: Made change.

api/src/services/QueueManager.ts
diboy2: Made change.

demiankatz (approved):

PR #97, opened 2021-11-29
Refactor paginator controls. Added Paginator Context

diboy2:


client/src/PaginatorContext.js
```diff
@@ -11,17 +12,61 @@ import {
...
+ * Update the shared states of react components.
+ */
+const paginatorReducer = (state, { type, payload }) => {
+    if (Object.keys(typePayloadMapping).includes(type)) {
+        return {
+            ...state,
+            [`${typePayloadMapping[type]}`]: payload,
+        };
+    } else {
+        console.error("action type does not exist", type);
```
demiankatz: Would this be better as `` `action "${type}" does not exist` ``?

client/src/PaginatorContext.js
```diff
@@ -75,24 +123,48 @@ const useJobPaginator = (initialCategory, initialJob) => {
...
             alert(message.join());
         }
     };
 
-    const loadJob = async () => {
-        const { order } = await fetchJSON(getJobUrl(category, job));
-        setOrder(order);
-        setCurrentPage(0);
-        updatePagesByStatus(await fetchJSON(getJobUrl(category, job, "/status")));
+    const setInitials = (initialCategory, initialJob) => {
```
demiankatz: I think the function name `setInitials` might be a little bit misleading/confusing. Could we just call this `initialize`?

client/src/PaginatorContext.js
```diff
@@ -32,12 +77,12 @@ const useJobPaginator = (initialCategory, initialJob) => {
     };
 
     const setLabel = (imageNumber, text) => {
-        setMagicLabelCache([]); // clear label cache whenever there is a change
+        dispatch({ type: "UPDATE_MAGIC_LABEL_CACHE", payload: [] }); // clear label cache whenever there is a change
```
demiankatz: I think in some (maybe all?) cases, it may be better to use setter methods that wrap around dispatch instead of calling dispatch directly, just to improve the readability of the code (it takes a bit more mental energy to parse out what a dispatch call is doing, vs. the instant comprehensibility of a setter). There may be some situations where creating a separate setter would just add unnecessary verbosity, but at very least, I think it would make sense to define setOrder early as a function and then use it in all the places where we change the order. Right now, we're changing order with dispatch in quite a few places, and the setter function is being defined inline at the very bottom of the file so that it can be used externally. I think it would probably reduce quite a few lines of code if we defined it earlier, used it in place of dispatch calls, and then exported the already-defined function.

client/src/PaginatorContext.js
```diff
@@ -75,24 +123,48 @@ const useJobPaginator = (initialCategory, initialJob) => {
...
+        });
+        dispatch({
+            type: "UPDATE_JOB",
+            payload: initialJob,
+        });
+    };
+
+    const loadJob = async (initialCategory, initialJob) => {
+        setInitials(initialCategory, initialJob);
+        const { order } = await fetchJSON(getJobUrl(initialCategory, initialJob));
```
demiankatz, line 173: What happens if this API call fails? Do we need any error handling here, or are things handled upstream? I wonder if nothing else if we should catch errors, pop up an alert dialog with details, and then either set the order to empty or somehow disable the whole component. (This may be an existing problem that should be handled in a separate PR/task, but it just came to mind while reviewing the code).

client/src/Thumbnail.test.js
```diff
@@ -28,21 +47,21 @@ describe("Thumbnail", () => {
         const wrapper = mount(<Thumbnail {...props} />);
 
         expect(props.scrollTo).toHaveBeenCalled();
-        expect(props.getJobImageUrl).toHaveBeenCalledWith(props.number, "thumb");
+        expect(paginatorValues.action.getJobImageUrl).toHaveBeenCalledWith(paginatorValues.state.order[0], "thumb");
```
demiankatz: Should you use props.number in place of the hard-coded 0 here? Should we add another test to cover a situation where there are multiple images? I wonder if testing the "one image only" edge case may not be sufficient on its own.

demiankatz: (There are also some more hard-coded 0's below that might be best replaced with props.number).

client/src/usePaginatorControls.js
```diff
@@ -0,0 +1,68 @@
+import { useState } from "react";
+import MagicLabeler from "./MagicLabeler";
+
+const usePaginatorControls = (currentPage, getMagicLabel, setLabel) => {
+    const [labelInput, setLabelInput] = useState("");
+    const getControlsLabel = (useMagic) => {
+        if (typeof useMagic === "undefined") {
```
demiankatz, line 7: Could we do null coalescing below (i.e. replace `useMagic` with `(useMagic ?? true)`) to simplify this code?

client/src/usePaginatorControls.test.js
```diff
@@ -0,0 +1,121 @@
...
+                result.current.action.toggleRoman("test");
+            });
+
+            expect(toggleRomanSpy).toHaveBeenCalledWith("testMagicLabel");
+            expect(setLabel).toHaveBeenCalledWith(currentPage, "testLabel");
+        });
+    });
+
+    describe("updateCurrentPageLabel", () => {
+        it("toggles the label to roman numerals", async () => {
```
demiankatz: This seems to be a copy and paste of the previous test; am I missing something?

It might be good to have a test for the edge case where toggleRoman returns false.

client/src/usePaginatorControls.test.js
```diff
@@ -0,0 +1,121 @@
...
```
demiankatz: The `it()` description here seems incorrect.

demiankatz (changes requested):
Thanks, @diboy2 -- once again, this looks really good, and significantly improves the code. I haven't done hands-on testing yet, but I'll do so after you have a chance to look at the feedback I'm providing here. Nothing too serious, but a few places where further minor improvements may be possible.

client/src/usePaginatorControls.test.js
diboy2: yeah sorry.

client/src/PaginatorContext.js
diboy2: made change

client/src/PaginatorContext.js
diboy2: made change

client/src/PaginatorContext.js
diboy2: made change

client/src/Thumbnail.test.js
diboy2: made change. added 3 objects.

client/src/usePaginatorControls.js
diboy2: From what I remember, this behaves differently whether it is undefined or false.

client/src/PaginatorContext.js
diboy2: At this moment, error handling is handled at the fetchJSON. But, I do think it needs to be refactored to be consistent on all calls. I remember talking about how the error handling was inconsistent and we probably need a task for the refactor.

client/src/usePaginatorControls.test.js
diboy2: made change. added false scenario.

client/src/usePaginatorControls.js
demiankatz: I think there is different behavior depending on whether the label is false or undefined, but this particular code is just designed to default the useMagic flag parameter to true if it is not provided (probably because Javascript didn't support default parameter values when I first wrote this code many years ago).

client/src/usePaginatorControls.test.js
demiankatz: I think this still needs to be revised -- maybe `it("sets the label of the current page")`

client/src/usePaginatorControls.test.js
demiankatz: Hmm, I'm not seeing the false scenario either. Am I missing something, or did your changes to this file somehow not get committed/pushed?

client/src/usePaginatorControls.test.js
demiankatz: Never mind, seeing it now. :-)

client/src/usePaginatorControls.test.js
demiankatz: ...and I must have been missing something, because all is well now.

client/src/PaginatorContext.js
demiankatz: Yes, agreed, error handling is inconsistent. I think it's reasonable to set this up so that fetchJSON throws an exception when something goes wrong... but then the calling code should trap that exception and do something with it, or else we'll just get unhandled exceptions blowing up the React environment (if I remember correctly how this all works). As you say, though, probably worth treating that as a separate task.

client/src/PaginatorContext.js
```diff
@@ -30,7 +75,9 @@ const useJobPaginator = (initialCategory, initialJob) => {
             }
         });
     };
-
+    const setOrder = (order) => dispatch({ type: "UPDATE_ORDER", payload: order });
```
demiankatz: Just a style question: you implemented setOrder and setMagicLabelCache as one-liners, but all the subsequent setters are formatted to take up several lines. Any reason for not using the same approach everywhere? Might be worth picking one or the other and being consistent.

diboy2: I think ill do it in several lines.

demiankatz (approved):
Just did some hands-on testing and this seems to be working fine!

PR #98, opened 2021-12-16
Eliminate use of RELS-EXT datastream

demiankatz:
With Fedora 6.1, the migration tool loads RDF from the RELS-EXT datastream directly into the new Fedora objects, so there is no longer a need to parse and manipulate legacy RELS-EXT XML binaries. This pull request implements the change.

Note that one significant data change is that `info:fedora/` URIs from RELS-EXT are transformed to use the actual base URI of the Fedora repository in place of `info:fedora/`.

TODO:
- [x] Update Solr indexer to pull from RDF instead of RELS-EXT
- [x] Modify other internal code that reads directly from RELS-EXT to use RDF instead
- [x] Modify code that writes to RELS-EXT to modify object RDF instead
- [x] Double-check for unwanted `info:fedora` references in code
- [x] Double-check for stray `RELS-EXT` or `relsext.` references in code
- [x] Revise FedoraObject::getSort() to use RDF instead of RELS-EXT
- [x] Add tests
- [x] Double-check that Tika extraction still works in refactored code
- [x] Double-check that external VuFind code referencing relsext fields is not impacted by the "info:fedora" change described above
  - [x] Fix getClasses() implementation
  - [x] Fix values in relsext.hasModel_txt_mv field for backward compatibility


Geoffsc:
Looking great so far, thanks for this!

demiankatz:
I spent some time today adding test coverage. I haven't added tests for two of the large service classes that are not already covered: HierarchyCollector and SolrIndexer. I expect some near-future refactoring of those classes, and I'd prefer to add tests as part of that refactoring rather than creating a lot of things that will need to change almost immediately. I think I've covered most of the other important details, though, and this PR should leave us with better coverage than we had before.

demiankatz:
I could probably still stand to add some test coverage to the edit route. I'll try to do that and also finish up the VuFind testing tomorrow, and then I think this may be ready for final review.

demiankatz:
I finished up testing today with coverage of the updated edit route.

I still need to finish VuFind testing, which I will try to fit in next week.

We should also merge #116 prior to this PR, since it will prevent a lot of noise in the logs.

demiankatz:
Okay, I believe that this work is finally done. I'd be interested in a review from @diboy2 and/or @crhallberg so that we can get this merged (along with #116, which is a prerequisite for this).

api/src/services/HierarchyCollector.ts
```diff
@@ -78,21 +73,19 @@ class HierarchyCollector {
...
-    async getHierarchy(pid: string, fetchRdf = true): Promise<FedoraData> {
-        const result = await this.getFedoraData(pid, fetchRdf);
+    async getHierarchy(pid: string): Promise<FedoraData> {
+        const result = await this.getFedoraData(pid);
         // Create promises to retrieve parents asynchronously...
-        const promises = (result.relations.isMemberOf ?? []).map(async (resource) => {
-            const parentPid = resource.substr("info:fedora/".length);
-            // The "false" here skips RDF retrieval:
-            const parent = await this.getHierarchy(parentPid, false);
+        const promises = (result.fedoraDetails.isMemberOf ?? []).map(async (resource) => {
```
diboy2, line 86: What happens when one of these promises fail? Do you have a try catch for this Promise.all?

demiankatz: At this point, all the error handling is occurring in the Express.js routes -- all calls (indirect or direct) to getHierarchy are wrapped in try..catch at the route layer so that appropriate error codes can be returned from the framework if something goes wrong.

api/src/services/SolrIndexer.ts
```diff
@@ -281,14 +275,26 @@ class SolrIndexer {
...
             }
         }
-
-        fields.has_order_str = ((fields["relsext.sortOn_txt_mv"] ?? [])[0] ?? "title") === "custom" ? "yes" : "no";
-
+        // Fedora 3 stored some data in Fedora object XML and some in separate RELS-EXT datastreams.
+        // Legacy VuDL indexed these two data sources using different prefixes. For stability of legacy
+        // queries, we retain this prefix separation, using the table below to identify the former
+        // RELS-EXT fields and defaulting to the "fgs." prefix for everything else.
+        const prefixes = {
```
diboy2, line 282: you can maybe reverse this mapping to do "relsext => ["hasLegacyUrl", "hasModal", ...] "

demiankatz: I used this mapping to keep the logic simple. If I reverse the map, then the logic gets more complicated. I took a stab at it to see what it would look like, but it's actually the same number of lines this way:

<pre>
         // queries, we retain this prefix separation, using the table below to identify the former
         // RELS-EXT fields and defaulting to the "fgs." prefix for everything else.
         const prefixes = {
-            hasLegacyURL: "relsext",
-            hasModel: "relsext",
-            itemID: "relsext",
-            isMemberOf: "relsext",
-            sequence: "relsext",
-            sortOn: "relsext",
+            relsext: ["hasLegacyURL", "hasModel", "itemID", "isMemberOf", "sequence", "sortOn"],
         };
         for (const field in fedoraData.fedoraDetails) {
-            const prefix = prefixes[field] ?? "fgs";
+            let prefix = "fgs";
+            for (const key in prefixes) {
+                if (prefixes[key].indexOf(field) !== -1) {
+                    prefix = key;
+                }
+            };
             const fieldName = prefix + "." + field + "_txt_mv";
             fields[fieldName] = fedoraData.fedoraDetails[field];
         }
</pre>

I'm fine with committing this version if you like it better... or if you have suggestions for making it more efficient!

crhallberg (approved):

api/src/services/SolrIndexer.ts
diboy2: okay.

api/src/services/HierarchyCollector.ts
diboy2: Okay.

diboy2 (approved):

demiankatz:
Thanks for the reviews, @crhallberg and @diboy2!

PR #99, opened 2021-12-20
Nextjs Refactor

diboy2:


client/.prettierignore
```diff
@@ -0,0 +1,6 @@
+# Ignore images
+*.svg
+
+# Ignore vender css
+client/src/css/justgrid.css
```
demiankatz: Would it make sense to create a `client/src/css/vendor` directory to isolate vendor CSS from local CSS in a more explicit way (and make it easier to drop in more if we need to)?

Also, if I'm not mistaken, justgrid.css was actually developed by @crhallberg (see [https://crhallberg.com/justgrid/](https://crhallberg.com/justgrid/)). I wonder if we should add a header comment to the file to indicate its source/license.

client/components/paginate/MainMenu.tsx
```diff
@@ -9,22 +8,22 @@ const MainMenu = () => {
...
                 <li>
-                    <Link to="/edit">Object Editor</Link>
+                    <Link href="/edit">Object Editor</Link>
                 </li>
             </ul>
             <h2 className="main-menu-other-tools-heading">Other Tools</h2>
             <ul className="main-menu-other-tools-list">
                 <li>
-                    <a href={baseUrl + "/queue"}>Arena (Job Queue Manager)</a>
+                    <a href={"/queue"}>Arena (Job Queue Manager)</a>
```
demiankatz: I think this introduces a bug: Arena is mounted as part of the back-end Express app, not the front-end React app. Thus, the link needs to point at the back-end URL.

client/pages/_app.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+import "../styles/normalize.css";
+import "../styles/application.css";
+import "../styles/client.css";
+import "../styles/justgrid.css";
+import "@fortawesome/fontawesome-free/css/all.min.css";
+
+function MyApp({ Component, pageProps }) {
+    return (
+        <PaginatorContextProvider>
```
demiankatz, line 14: Does it make sense to put the PaginatorContextProvider around the whole application, or would it be better to put this only around the Paginator itself? (If this keeps things simpler, I don't mind it... but it seems potentially over-broad).

package.json
```diff
@@ -16,7 +16,7 @@
     "api:testWatch": "cd api && npm run testWatch",
     "api:watch": "cd api && tsc --watch",
     "api:wsl": "nodemon --legacy-watch api/bin/www",
-    "client": "cd client && npm run start",
+    "client": "cd client && npm run dev",
```
demiankatz: Would it make sense to keep the default client task as `start` instead of `dev` and then create a separate `client:dev` script which is called by the top-level `dev` script? It seems like this might be a little more predictable/consistent.

demiankatz (commented):
Thanks, @diboy2, for the great progress here. I really like how this has shaped up! I've just left a handful of comments based on an initial review. I'll plan to do a deeper, hands-on test in the new year, once the build is fully passing here.

client/.prettierignore
diboy2: Sorry, this was copy/paste.

demiankatz:
@diboy2, I took a little bit of time to test this today (though nothing exhaustive yet). I decided to try to deploy it on Node.js v14, since that may be the version we'll be limited to in production based on OS distributions. It seems to work, but I had to delete package-lock.json and rerun composer install before it would work right. That's the reason for commit fde875d; please let me know if that commit has messed anything up for your environment.

demiankatz:
(I'll take some more time to test more exhaustively once you've had a chance to address the comments above).

client/.prettierignore
diboy2: Made change

client/components/paginate/MainMenu.tsx
diboy2: Made change.

diboy2: Made change.

client/pages/_app.tsx
diboy2: Made change

diboy2: So after further review, we made this change because some paginator components are rendered by route, so this the topmost level where the paginator states are shared.

demiankatz: Ahh, I see, sorry for the false alarm. :-)

package.json
diboy2: Made change

client/package.json
```diff
@@ -1,58 +1,43 @@
 {
-  "name": "client",
-  "version": "0.1.0",
+  "name": "vudl-test",
```
demiankatz: Should we keep the name as client? Or perhaps use vudl-client? I don't think "test" is appropriate at this stage.

package.json
```diff
@@ -16,9 +16,10 @@
     "api:testWatch": "cd api && npm run testWatch",
     "api:watch": "cd api && tsc --watch",
     "api:wsl": "nodemon --legacy-watch api/bin/www",
-    "client": "cd client && npm run start",
```
demiankatz, line 19: Should we put back the "client" script but make it run the "prod" instead of "start" script? I think this would be consistent with the current "backend" script, which runs the backend in production mode.

client/public/vercel.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="283" height="64" viewBox="0 0 283 64" fill="none"
```
demiankatz: Do we actually need this file? Is it being displayed anywhere, or just part of the boilerplate?

demiankatz (commented):
Thanks for the progress on this today -- a few more minor comments from my latest review of the whole PR.

Regarding the test issue I alluded to in our chat, I'm still seeing this warning while running the client test suite:

<pre>
   console.error
      Warning: <mock-JobClickable /> is using incorrect casing. Use PascalCase for React components, or lowercase for HTML elements.
          at mock-JobClickable
          at JobClickable
          at li
          at Job (/opt/vudlprep/client/components/paginate/Job.tsx:7:16)
          at FetchContextProvider (/opt/vudlprep/client/context/FetchContext.tsx:31:40)

      at printWarning (node_modules/react-dom/cjs/react-dom-server.node.development.js:82:30)
      at error (node_modules/react-dom/cjs/react-dom-server.node.development.js:58:5)
      at ReactDOMServerRenderer.renderDOM (node_modules/react-dom/cjs/react-dom-server.node.development.js:4072:11)
      at ReactDOMServerRenderer.render (node_modules/react-dom/cjs/react-dom-server.node.development.js:3802:21)
      at ReactDOMServerRenderer.read (node_modules/react-dom/cjs/react-dom-server.node.development.js:3690:29)
      at Object.renderToStaticMarkup (node_modules/react-dom/cjs/react-dom-server.node.development.js:4314:27)
      at Object.render (node_modules/@wojtekmaj/enzyme-adapter-react-17/src/ReactSeventeenAdapter.js:782:31)
      at render (node_modules/enzyme/src/render.js:21:25)
</pre>

...not sure if there's an easy way to fix it, or if it's just an unavoidable artifact of how the tests are set up.

package.json
diboy2: okay.

client/public/vercel.svg
diboy2: I think it's part of the boilerplate. I missed that.

client/package.json
diboy2: Made chagne.

diboy2:
@demiankatz fixed that pascal case issue.

client/public/vercel.svg
diboy2: Removed files, but kept public folder because I think it's useful to store image files such as brand.

package.json
diboy2: Made change.

client/public/vercel.svg
demiankatz: Makes sense, thanks!

client/components/edit/CreateObject.test.tsx
```diff
@@ -5,8 +5,8 @@ import { waitFor } from "@testing-library/react";
 import { mount, render } from "enzyme";
 import toJson from "enzyme-to-json";
 import CreateObject from "./CreateObject";
-import { FetchContextProvider } from "../FetchContext";
-import { BrowserRouter } from "react-router-dom";
+import { FetchContextProvider } from "../../context/FetchContext";
+// import { BrowserRouter } from "react-router-dom";
```
demiankatz: If we don't need the BrowserRouter, should we remove the commented-out lines at this point?

package.json
```diff
@@ -16,9 +16,11 @@
...
+    "client": "cd client && npm run pro d",
```
demiankatz: I think there's a typo here - `pro d` should be `prod`.

demiankatz (changes requested):
Thanks! We're very nearly there, but I noticed a couple of very small things while reviewing the latest changes. I'll do some hands-on testing now and let you know if anything breaks in my environment.

demiankatz:
@diboy2, one other observation: in running the tests, I see that we have no tests for the JobClickable or PaginatorControls components. We should add those at some point; possibly something we could do as part of this project, but if you prefer to get this merged sooner rather than later, maybe we should just add this as a future task to address separately.

demiankatz:
@diboy2, I've run into an interesting problem; perhaps you have an idea of the cause. Steps to reproduce:

1.) Go into the job paginator, and select a job.
2.) Save the job and acknowledge any pop-up messages.
3.) You will now be returned to the job paginator as part of the Save process.
4.) Hit the back button -- this takes you back to the job you were previously editing.
5.) You now have a runtime error on screen (or at least, I consistently do in my environment):

runtime-error

I haven't tried testing this in the old environment to see if something similar happens, but I suspect this is somehow related to the way Next.js manages state.

client/components/edit/CreateObject.test.tsx
demiankatz: I see you got rid of the import, but there are also a couple other BrowserRouter-related commented lines below that should also go.

demiankatz (approved):
This all looks good to me now -- apart from the one remaining comment about cleaning up some unneeded commented-out lines, which I imagine you can take care of in a matter of seconds, I think this is ready to merge. :-)

crhallberg (approved):

client/components/edit/CreateObject.test.tsx
diboy2: made change

PR #100, opened 2022-01-13
Add datastream editor

demiankatz:
TODO
- [ ] Add logic/configuration to list allowed datastreams along with actual datastreams
- [ ] Actually implement editing capabilities
- [ ] Update with changes from #99 
- [ ] Write tests


demiankatz:
As of right now, this is just a stub proof-of-concept for the datastream editor: a new API endpoint that provides models and existing datastreams for a given PID, and a React component that displays these data and which is integrated into the existing object editor component. Hopefully this provides a simple foundation that we can build upon.

demiankatz:
Note that I'm not especially happy with the URL patterns for our object-related API routes, but I'm not worrying about improving that right now since we have several open PRs that are adding different pieces. It's probably easier to clean this up after everything has been merged into one place.

api/src/routes/edit.ts
```diff
@@ -81,6 +82,15 @@ async function getChildren(req, res) {
     res.json(response);
 }
 
+edit.get("/object/details/:pid", requireToken, pidSanitizer, async function (req, res) {
+    try {
+        const data = await HierarchyCollector.getInstance().getFedoraData(req.params.pid);
```
demiankatz, line 87: HierarchyCollector is really a bad name for this service. We should probably rename it to something like FedoraLookup. Again, probably a problem to solve in the future... but just mentioning it because it's starting to bug me.

demiankatz:
Superseded by #111.

PR #101, opened 2022-01-14
Fix audit issues for sqlite3.

demiankatz:
A few months ago, we struggled to get `npm audit` to pass. The problem persists, and the root cause appears to be the fact that the sqlite library that we use hasn't issued a release in a while. There's a fix to the problem in their Git repo, however, and pointing directly to a commit fixes everything.

This PR applies that change. It would be better to have an official release, but we can decide whether or not to use this workaround in the short term. Either way, this PR is a reminder to resolve this issue.

(Solution taken from https://github.com/mapbox/node-sqlite3/issues/1493#issuecomment-927877859)

### DIFF
--- a/api/package-lock.json
+++ b/api/package-lock.json
@@ -36,7 +36,7 @@
         "passport": "^0.4.1",
         "passport-hash": "^0.6.1",
         "pdfkit": "^0.12.1",
-        "sqlite3": "^5.0.2",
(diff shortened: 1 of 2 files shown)


demiankatz:
Here are threads discussing this issue:

https://github.com/mapbox/node-sqlite3/issues/1483

https://github.com/mapbox/node-sqlite3/issues/1493

demiankatz:
A new release has finally been issued, so this ugly workaround is no longer needed. Superseded by #130.

PR #102, opened 2022-01-14
Remove api:watch npm script.

demiankatz:
The `api:watch` NPM script was set up to recompile the API code upon changes, and used to need to be run in parallel with the `dev` script to have a running dev environment that would automatically update itself upon code changes. Since @diboy2 has streamlined some configurations, the `dev` task now handles watching for changes by itself, and `api:watch` is no longer needed (in fact, it currently causes tests to break by generating an unneeded `dist` directory).

This PR removes the obsolete script.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,6 @@ After a few moments, a new tab should automatically open in your browser pointin
 | api:dev | run api server, restart server on changes |
 | api:lint | lint only the api code |
 | api:setup | install npm dependencies in api/ |
-| api:watch | run Typescript in watch mode on api code |
 | api:wsl | run api server, restart on changes (use if on WSL or `api:dev` isn't working) |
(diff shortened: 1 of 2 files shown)


diboy2 (approved):

crhallberg (approved):
You can probably remove the wsl scripts too after the Next migration.

demiankatz:
Thanks, @crhallberg. I'll go ahead and merge this now. Regarding the wsl scripts, I agree that they should probably be removed, though note that the api piece is still using Express, not Next.js; I think @diboy2 just tweaked some configurations on the API side. Do you want to test again when time permits to see if those can be removed? (Even if we need to keep them, I think some changes might be in order... but hopefully we can just get rid of them to prevent future confusion).

PR #103, opened 2022-01-20
Return logout button

diboy2:


demiankatz (approved):
Looks good! Merging now. Thanks, @diboy2!

PR #104, opened 2022-01-21
Add object details.

demiankatz:
This PR is intended to add an "object details" display and corresponding API.

TODO
- [x] Add more data elements to API output
- [x] Create React component to display title, description and thumbnail
- [x] Incorporate new component into object editor
- [x] Write tests for new component

### DIFF
--- a/api/src/routes/edit.ts
+++ b/api/src/routes/edit.ts
@@ -2,6 +2,7 @@ import express = require("express");
 import bodyParser = require("body-parser");
 import Config from "../models/Config";
 import Fedora from "../services/Fedora";
+import { FedoraObject } from "../models/FedoraObject";
 import FedoraObjectFactory from "../services/FedoraObjectFactory";
 import MetadataExtractor from "../services/MetadataExtractor";
 import { requireToken } from "./auth";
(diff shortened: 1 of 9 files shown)


demiankatz:
While working on #98, I created a tiny API endpoint to test the revised sorting code. Rather than throwing that work away, I created this PR as a stub for some other functionality we need, which this API might be a starting point for. I plan to work on this later as time permits.

demiankatz:
I believe this is ready to merge now -- it adds a box at the top of the object editor showing the current object's title and description (and we should add thumbnail in future, once the datastream work is done). It also creates a new API route to feed the component, which includes a few additional data elements that we may be interested in using in future components (or perhaps we'll decide to revise/split/parameterize the API... just keeping it simple for now).

Geoffsc:
This looks good to me, thanks!

api/src/routes/edit.ts
```diff
@@ -83,5 +84,13 @@ async function getChildren(req, res) {
 
 edit.get("/object/children", requireToken, getChildren);
 edit.get("/object/children/:pid", requireToken, pidSanitizer, getChildren);
+edit.get("/object/details/:pid", requireToken, pidSanitizer, async function (req, res) {
+    const pid = req.params.pid;
+    const obj = FedoraObject.build(pid);
+    const sort = await obj.getSort();
+    const metadata = await Fedora.getInstance().getDC(pid);
```
diboy2, line 91: What does DC mean? Is it possible to have a more descriptive function?

demiankatz: DC stands for [Dublin Core](https://dublincore.org/), the metadata standard used by the digital library. I wouldn't object to renaming this method to getDublinCore for clarity if you wish, though that's probably best handled in a separate PR.

diboy2: Okay.

diboy2 (approved):

PR #105, opened 2022-01-21
Model Tests

Geoffsc:


api/src/models/AudioFile.test.ts
```diff
@@ -0,0 +1,21 @@
...
+import Config from "../models/Config";
+
+jest.mock("../models/Config");
+
+describe("Audio", () => {
+    let audio: AudioFile;
+    let config;
+    beforeEach(() => {
+        config = {
+            ffmpegPath: "/usr/bin/ffmpeg"
```
demiankatz: It might be better to use a less realistic value here, like `/foo/ffmpeg`, so that if the test somehow triggers real behavior, it doesn't accidentally run a process that we don't really want to run. Or it might be better not to set this at all in the `beforeEach` but wait until we're testing a method that requires the setting, and add it as part of the test setup. (Whether to do this sort of thing globally or locally is kind of a judgment call -- depends on how many tests rely on the same input, etc.).

api/src/models/Category.test.ts
```diff
@@ -0,0 +1,19 @@
...
+    let category: Category;
+    beforeEach(() => {
+        category = new Category("test1");
+    });
+
+    it("should return the directory name", () => {
+        expect(category.dir).toEqual("test1");
+    });
+
+    it("should return the category and jobs", () => {
```
demiankatz: Maybe add "...in raw format" to the end of the test description, since this is mainly about testing the raw() method.

api/src/models/ImageFile.test.ts
```diff
@@ -0,0 +1,22 @@
...
+import Config from "../models/Config";
+
+jest.mock("../models/Config");
+
+describe("Image", () => {
+    let image: ImageFile;
+    let config;
+    beforeEach(() => {
+        config = {
+            tesseractPath: '/usr/local/bin/tesseract',
```
demiankatz: Same comment as above for these file paths.

api/src/models/Job.test.ts
```diff
@@ -0,0 +1,27 @@
...
+jest.mock("../services/QueueManager");
+
+describe("Job", () => {
+    let job: Job;
+    let config;
+    let queueManager;
+    beforeEach(() => {
+        config = { };
+        queueManager = { 
+            performIndexOperation: jest.fn(),
```
demiankatz: I wouldn't bother mocking these methods until we actually have a test that is going to call them.

demiankatz (commented):
Thanks, @Geoffsc -- just a few suggestions!

api/src/models/AudioFile.test.ts
Geoffsc: fixed

api/src/models/Category.test.ts
Geoffsc: fixed

api/src/models/ImageFile.test.ts
Geoffsc: fixed

api/src/models/Job.test.ts
Geoffsc: fixed

demiankatz:
Thanks for the updates, @Geoffsc! I think we could merge this, though it might be good to do one more thing before we do: since you've set up mocks and fake configs in a few tests, but the tests don't exercise any of that logic or configuration, that could potentially confuse us in the future. It might be good to add tests that actually use that configuration so that its purpose is clear (or, alternatively, to comment it out until we need it). I'm fine with either approach, and we can discuss further later today!

demiankatz:
@Geoffsc, I think you got a little carried away with the commenting, and you commented out stuff that actually is used by the tests. The `beforeEach` methods at least need to set up the objects being tested -- I just thought it would be best not to set up configurations and mock objects before they are used by any of the code currently being covered by tests. Can you get these passing again? Let me know if you want to discuss!

demiankatz:
@Geoffsc, I'm seeing the same errors from continuous integration on my local test instance -- strange that it says the tests pass but returns an error code, but that seems to be what's happening. I don't have time to dig into this yet, but [this StackOverflow](https://stackoverflow.com/questions/50793885/referenceerror-you-are-trying-to-import-a-file-after-the-jest-environment-has) might shed some light. It might be useful to try deleting the new tests and reintroducing them one by one to see if there's a particular one that's causing the error to happen (the reference error is not occurring on the dev branch, so it's somehow related to something that changes here... though it's not obvious at a glance what or why).

demiankatz:
Since @Geoffsc has departed, I don't expect more work to be done on this pull request, so I polished up what was already here and got it merged. It only ticks up our test coverage a little bit, but progress is progress, and we can flesh these tests out further as time permits.

PR #107, opened 2022-01-27
Turn on next.js ESLint rules.

demiankatz:
This pull request turns on next.js ESLint rules.

I've disabled the "use Image instead of img" rule for now because I think it adds complexity we don't necessarily need.

I spent nearly an hour trying to figure out how to resolve `react-hooks/exhaustive-deps`, but I clearly don't understand all of the implications, because every time I make lint happy, I actually break the code and cause infinite AJAX reloads and similar problems. I get the impression that this is actually highlighting an important issue that we need to clean up, but I don't think my understanding of hooks is deep enough to actually fix it properly. Any advice, @diboy2?

(For the record, my first attempted solution was to simply add all the missing dependencies as suggested... but that clearly causes effects to trigger more frequently than we want them to; I next tried incorporating useCallback, but that didn't help either...).

TODO:
- [x] Resolve outstanding issues.

### DIFF
--- a/client/.eslintrc.json
+++ b/client/.eslintrc.json
@@ -18,7 +18,8 @@
       {
         "usePrettierrc": true
       }
-    ]
+    ],
+    "react-hooks/exhaustive-deps": "off"
   },
   "settings": {
     "react": {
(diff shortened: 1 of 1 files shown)


demiankatz:
As we just discussed, I have turned off the exhaustive-deps rule for now since we haven't found a clean way to solve it, and we can open a ticket to revisit this when we have more time.

PR #108, opened 2022-02-02
use "DublinCore" instead of DC

Geoffsc:


demiankatz:
Thanks, @Geoffsc! Merging now!

PR #109, opened 2022-02-04
Add ability to upload datastream file from api side

diboy2:


api/vudl.ini.dist
```diff
@@ -194,3 +194,118 @@ DOC = "vudl-system:DOCData"
...
+allowedType = "audio"
+allowedSubtypes = "mpeg3,x-mpeg-3,mpeg"
+
+[datastreams.MP3.mimetype]
+allowedType = "audio"
+allowedSubtypes = "ogg"
+
+# CollectionModel
+
+[datastreams.MEMBER-QUERY.mimetype]
```
demiankatz: I don't think we need to keep configuration for deprecated/legacy datastreams in the default configuration. Things like MEMBER-QUERY, MEMBER-LIST-RAW, STRUCTMAP, etc., etc. will never be added to the system going forward, so I don't think we need to specify legal types. Hopefully that will significantly reduce the size of this configuration.

api/src/models/FedoraData.ts
```diff
@@ -1,9 +1,17 @@
+export interface FedoraDatastream {
+    stream: string;
+    mimetype?: {
```
demiankatz: I'm not convinced that FedoraData is the right place to put the allowedType/allowedSubtypes settings. FedoraData is for retrieving the information that is already stored in Fedora. Adding validation rules to that feels to me like it belongs in higher-level code, rather than incorporated into the data retrieval process.

I particularly feel this way because the set of datastreams ALREADY stored in Fedora is not the same as the set of datastreams ALLOWED to be stored in Fedora based on the model-to-datastream rules we have established in Geoff's spreadsheet. I think perhaps that is the piece that still needs to be implemented here: code that takes a list of models and returns a list of allowed datastreams.

In other words, rather than retrieving a list of datastreams from Fedora and then merging in mime type configurations, maybe we need a new class which takes as input configuration, a list of datastreams from Fedora, and a list of models from Fedora. Given this, it creates a comprehensive list of legal datastreams and mimetypes, and it can also flag which datastreams already have content. We can do that extra work when we need to (as part of the editing interface) but then we don't need to do extra mapping/config-reading in contexts where the constraints don't matter (e.g. when we are indexing the existing data in the system but have no intention of writing or validating anything).

api/src/routes/edit.ts
```diff
@@ -82,6 +89,37 @@ async function getChildren(req, res) {
...
+edit.get("/object/details/:pid", requireToken, pidSanitizer, async function (req, res) {
+    try {
+        const data = await HierarchyCollector.getInstance().getFedoraData(req.params.pid);
+        res.json({ models: data.models, datastreams: data.fedoraDatastreams });
+    } catch (error) {
+        console.error(error);
+        res.status(500).send(error.message);
+    }
+});
+edit.post("/object/:pid", requireToken, pidSanitizer, uploadFile);
```
demiankatz, line 135: Maybe we should use a more specific path here, like `/object/datastreamUpload/:pid`. In theory, `:pid` can be practically anything (though in practice, we constrain it). I want to be sure we avoid ambiguity in our URL pathing.

I should also say that I'm not very happy with the current `/object/[action]/[pid]` pattern, when I think `/object/[pid]/action` would make a lot more sense. It just got started along this road because of the way we set up `/object/children` to get all top-level objects in the hierarchy, with `/object/children/:pid` to get children of a specific object. With that precedent set, everything else fell into place... but I'd really like to change it. If we rename `/object/children` to something like `/hierarchy/top` then the rest could probably be flipped around... but there have always been so many pull requests in progress at once, there hasn't been a good opportunity to address this in a single place.

api/src/routes/edit.ts
```diff
@@ -82,6 +89,37 @@ async function getChildren(req, res) {
...
+            res.status(200).send("Upload success");
+        } catch (error) {
+            res.status(500).send(error.message);
+        }
+    });
+}
+
+edit.get("/object/details/:pid", requireToken, pidSanitizer, async function (req, res) {
+    try {
+        const data = await HierarchyCollector.getInstance().getFedoraData(req.params.pid);
```
demiankatz, line 128: This code needs to be reconciled with the work that was merged from #104.

demiankatz (changes requested):
Thanks, @diboy2, this is off to a good start, though I have a few significant suggestions for changes. It also looks like there are some lint issues and merge conflicts that need to be sorted out (though of course, lint can wait until the design is stabilized -- no point wasting time polishing things that are likely to change).

I think the biggest thing to think about is incorporating the model-to-datastream mappings that inform which datastreams are allowed, and possibly moving the mime type validation to a different layer (discussed in more detail below).

I realize some of these comments are a bit rambly. Happy to talk about this if anything is unclear or you want to bounce ideas back and forth!

api/src/routes/edit.ts
demiankatz: (It looks like the merge just created two different routes with the same path).

crhallberg (approved):
Looks good to me once Demian's comments are addressed.

api/vudl.ini.dist

api/src/models/FedoraData.ts
diboy2: Made change.

api/vudl.ini.dist
```diff
@@ -194,3 +194,118 @@ DOC = "vudl-system:DOCData"
 Audio = "vudl-system:AudioData"
 XLS = "vudl-system:XLSData"
 Text = "vudl-system:TextData"
+
+# AudioData
+[models]
+AudioData.datastreams.WAV.mimetype.allowedType = "audio"
```
demiankatz, line 200: Any particular reason why WAV is formatted differently than subsequent formats? Should we make a `[models.AudioData.datastreams.WAV.mimetype]` section for consistency with what we're doing below?

api/vudl.ini.dist
```diff
@@ -194,3 +194,118 @@ DOC = "vudl-system:DOCData"
...
+
+[models.DataModel.datastreams.MASTER-MD.mimetype]
+allowedType = "text"
+allowedSubtypes = "xml"
+
+# FolderCollection
+
+[models.FolderCollection.datastreams]
+
+# HTMLData
```
demiankatz, line 248: HTMLData was never used -- we can delete this.

api/vudl.ini.dist
```diff
@@ -194,3 +194,118 @@ DOC = "vudl-system:DOCData"
...
+
+[models.ImageData.datastreams.OCR-EDITED.mimetype]
+allowedType = "text"
+allowedSubtypes = "plain"
+
+# ListCollection
+
+[models.ListCollection.datastreams]
+
+# MP3Data
```
demiankatz, line 275: MP3Data and Model were also unused and can be deleted.

api/vudl.ini.dist
```diff
@@ -194,3 +194,118 @@ DOC = "vudl-system:DOCData"
...
+
+[models.ResourceCollection.datastreams.AGENTS.mimetype]
+allowedType = "text"
+allowedSubtypes = "xml"
+
+[models.ResourceCollection.datastreams.PROCESS-MD.mimetype]
+allowedType = "text"
+allowedSubtypes = "xml"
+
+# SegmentImage
```
demiankatz, line 297: SegmentImage, TextData and WAVData can also all be deleted -- never used.

api/src/services/SolrIndexer.ts
```diff
@@ -322,12 +322,12 @@ class SolrIndexer {
         }
 
         // Change tracker details:
-        const change = await this.getChangeTrackerDetails(
-            pid,
-            fields["fgs.lastModifiedDate_txt_mv"][0] ?? "1900-01-01T00:00:00Z"
-        );
-        fields.first_indexed = change.getFirstIndexed;
-        fields.last_indexed = change.getLastIndexed;
+        // const change = await this.getChangeTrackerDetails(
```
demiankatz, line 325: This change was made to work around the fact that you don't have a working VuFind instance in your test environment, but it should not be committed here; I suspect this was included by accident.

A better solution to this problem would be to adjust `getChangeTrackerDetails` so that if the VuFind URL is set to false in the configuration, it returns false or null, and then skipping these fields if VuFind is not configured. That would make it easier to test without a full VuFind, and also without requiring code-hacking to avoid errors. Maybe that's worth doing in a separate PR.

demiankatz (commented):
Thanks for the progress, @diboy2 -- I like the direction this is taking. I left a few more (mostly unrelated) comments.

api/vudl.ini.dist
diboy2: Sorry. that's a typo.

diboy2:
Closing and combining both api and client in a pull request

PR #111, opened 2022-02-14
Add ability to upload datastream files

diboy2:
Screenshot 2022-02-14 11 47 09 AM

### DIFF
--- a/api/src/routes/edit.ts
+++ b/api/src/routes/edit.ts
@@ -2,16 +2,36 @@ import express = require("express");
 import bodyParser = require("body-parser");
 import Config from "../models/Config";
 import Fedora from "../services/Fedora";
+import FedoraCatalog from "../services/FedoraCatalog";
 import { FedoraObject } from "../models/FedoraObject";
+import DatastreamManager from "../services/DatastreamManager";
 import FedoraObjectFactory from "../services/FedoraObjectFactory";
+import HierarchyCollector from "../services/HierarchyCollector";
 import MetadataExtractor from "../services/MetadataExtractor";
 import { requireToken } from "./auth";
 import { pidSanitizer } from "./sanitize";
+import * as formidable from "formidable";
 import Solr from "../services/Solr";
 const edit = express.Router();
 
 edit.get("/models", requireToken, function (req, res) {
     res.json({ CollectionModels: Config.getInstance().collectionModels, DataModels: Config.getInstance().dataModels });
+});
+
+edit.get("/catalog", requireToken, function (req, res) {
+    res.json({ models: FedoraCatalog.getInstance().getCompleteCatalog() });
+});
+
+edit.get("/catalog/models", requireToken, function (req, res) {
+    res.json(FedoraCatalog.getInstance().getModelCatalog());
+});
+
+edit.get("/catalog/datastreams", requireToken, function (req, res) {
+    res.json(FedoraCatalog.getInstance().getDatastreamCatalog());
+});
+
+edit.get("/catalog/datastreammimetypes", requireToken, function (req, res) {
+    res.json(FedoraCatalog.getInstance().getDatastreamMimetypes());
 });
 
 edit.post("/object/new", requireToken, bodyParser.json(), async function (req, res) {
@@ -82,6 +102,37 @@ async function getChildren(req, res) {
     res.json(response);
 }
 
+function uploadFile(req, res, next) {
+    const { pid } = req.params;
+    const form = formidable({ multiples: true });
+
+    form.parse(req, async (err, fields, files) => {
+        if (err) {
+            next(err);
+            return;
+        }
+        try {
+            const datastream = DatastreamManager.getInstance();
+            const { stream } = fields;
+            const { filepath, mimetype } = files?.file;
+            await datastream.uploadFile(pid, stream, filepath, mimetype);
+            res.status(200).send("Upload success");
+        } catch (error) {
+            res.status(500).send(error.message);
+        }
+    });
+}
+
+edit.get("/object/modelsdatastreams/:pid", requireToken, pidSanitizer, async function (req, res) {
+    try {
+        const data = await HierarchyCollector.getInstance().getFedoraData(req.params.pid);
+        res.json({ models: data.models, datastreams: data.fedoraDatastreams });
+    } catch (error) {
+        console.error(error);
+        res.status(500).send(error.message);
+    }
+});
+edit.post("/object/:pid", requireToken, pidSanitizer, uploadFile);
 edit.get("/object/children", requireToken, getChildren);
 edit.get("/object/children/:pid", requireToken, pidSanitizer, getChildren);
 edit.get("/object/details/:pid", requireToken, pidSanitizer, async function (req, res) {
--- a/api/src/services/FedoraCatalog.ts
+++ b/api/src/services/FedoraCatalog.ts
@@ -1,0 +1,51 @@
+import Config from "../models/Config";
+
+export interface FedoraDatastream {
+    mimetype?: {
+        allowedType: string;
+        allowedSubtypes: string;
+    };
+}
+
+export interface FedoraModel {
+    datastreams?: Record<string, FedoraDatastream>;
+}
+
+class FedoraCatalog {
+    private static instance: FedoraCatalog;
+
+    config: Config;
+
+    constructor(config: Config) {
+        this.config = config;
+    }
+
+    public static getInstance(): FedoraCatalog {
+        if (!FedoraCatalog.instance) {
+            FedoraCatalog.instance = new FedoraCatalog(Config.getInstance());
+        }
+        return FedoraCatalog.instance;
+    }
+
+    getCompleteCatalog(): Record<string, FedoraModel> {
+        return this.config.models;
+    }
+
+    getModelCatalog(): Array<string> {
+        return Object.keys(this.config.models);
+    }
+
+    getDatastreamCatalog(): Array<string> {
+        return Object.values(this.config.models).reduce((acc, model) => {
+            return [...acc, ...Object.keys(model.datastreams)];
+        }, []);
+    }
+
+    getDatastreamMimetypes(): Record<string, FedoraDatastream> {
+        return Object.values(this.config.models).reduce((acc: Record<string, FedoraDatastream>, model: FedoraModel) => {
+            return { ...acc, ...model.datastreams };
+        }, {});
+    }
+}
+
+export default FedoraCatalog;
--- a/api/src/services/HierarchyCollector.ts
+++ b/api/src/services/HierarchyCollector.ts
@@ -37,28 +37,45 @@ class HierarchyCollector {
         // we can skip fetching more RDF in order to save some time!
         const RDFPromise = fetchRdf ? this.fedora.getRdf(pid) : null;
         const [DC, RELS, RDF] = await Promise.all([DCPromise, RELSPromise, RDFPromise]);
-        const dataStreams = fetchRdf ? this.extractor.extractFedoraDatastreams(RDF) : [];
         const relations = this.extractor.extractRelations(RELS);
+        const datastreams = fetchRdf ? this.extractor.extractFedoraDatastreams(RDF) : [];
+        const extraDetails = await this.getExtraDetails(pid, datastreams, relations);
+
+        return new FedoraData(
+            pid,
+            relations,
+            this.extractor.extractMetadata(DC),
+            fetchRdf ? this.extractor.extractFedoraDetails(RDF) : {},
+            datastreams,
+            extraDetails
+        );
+    }
+
+    async getExtraDetails(
+        pid: string,
+        datastreams: Array<string>,
+        relations: Record<string, Array<string>>
+    ): Promise<Record<string, Record<string, Array<string>>>> {
         // Fetch license details if appropriate/available:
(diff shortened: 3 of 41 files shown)


api/src/routes/edit.ts
```diff
@@ -82,6 +102,37 @@ async function getChildren(req, res) {
...
+            const { filepath, mimetype } = files?.file;
+            await datastream.uploadFile(pid, stream, filepath, mimetype);
+            res.status(200).send("Upload success");
+        } catch (error) {
+            res.status(500).send(error.message);
+        }
+    });
+}
+
+edit.get("/object/details/:pid", requireToken, pidSanitizer, async function (req, res) {
```
demiankatz: This is a duplicate route name -- see line 87/138 below.

api/src/services/Datastream.test.ts
```diff
@@ -0,0 +1,95 @@
...
+
+        it("returns false when type does not exist", () => {
+            expect(datastream.hasValidMimeType("WAV", "test5/test2")).toBeFalsy();
+        });
+
+        it("returns true when type/subtype does exist", () => {
+            expect(datastream.hasValidMimeType("WAV", "test1/test2")).toBeTruthy();
+        });
+
+        it("returns true for any type", () => {
```
demiankatz: This appears to be a duplicate test case (combining code from above and name from below).

api/src/services/Datastream.ts
```diff
@@ -0,0 +1,39 @@
...
+        }
+        return Datastream.instance;
+    }
+
+    async uploadFile(pid: string, stream: string, filepath: string, mimeType: string): Promise<void> {
+        const fedoraObject = FedoraObject.build(pid);
+        if (this.hasValidMimeType(stream, mimeType)) {
+            await fedoraObject.updateDatastreamFromFile(filepath, stream, mimeType);
+        } else {
+            throw new Error("Invalid mime types");
```
demiankatz: Would it be worth saying `"Invalid mime type: " + mimeType` instead?

api/src/services/FedoraCatalog.ts
```diff
@@ -0,0 +1,51 @@
+import Config from "../models/Config";
+
+export interface FedoraDatastream {
+    mimetype?: {
+        allowedType: string;
```
demiankatz, line 5: I notice that allowedType is singular but allowedSubtypes is plural. It would probably make more sense to make them consistent, since they both work the same way -- I would vote for changing allowedType to allowedTypes.

api/src/services/HierarchyCollector.ts
```diff
@@ -37,43 +37,53 @@ class HierarchyCollector {
         // we can skip fetching more RDF in order to save some time!
         const RDFPromise = fetchRdf ? this.fedora.getRdf(pid) : null;
         const [DC, RELS, RDF] = await Promise.all([DCPromise, RELSPromise, RDFPromise]);
-        const dataStreams = fetchRdf ? this.extractor.extractFedoraDatastreams(RDF) : [];
         const relations = this.extractor.extractRelations(RELS);
+        const dataStreams = fetchRdf ? this.extractor.extractFedoraDatastreams(RDF) : [];
```
demiankatz: Since you changed `dataStreams` to `datastreams` in the code below, maybe it makes sense to also change it in this function for consistency (I agree that all lowercase seems more appropriate, given the way we use this word).

api/vudl.ini.dist
```diff
@@ -194,3 +194,93 @@ DOC = "vudl-system:DOCData"
 Audio = "vudl-system:AudioData"
 XLS = "vudl-system:XLSData"
 Text = "vudl-system:TextData"
+
+# AudioData
+[AudioData.datastreams.WAV.mimetype]
```
demiankatz: Should this have a `models` prefix on it to match other entries below?

demiankatz (commented):
Thanks for the progress on this! I still need to find time to do some hands-on testing, but here are a few observations (mostly backend-focused) based on an initial readthrough of the code.

crhallberg (approved):

demiankatz:
@diboy2, a couple more observations based on brief hands-on testing:

1.) When I forgot to add the new model data to my vudl.ini, it failed like this:

missing-config-error

Of course, that was my mistake, but maybe this is a sign we need a little more error-handling or validation somewhere.

2.) The object details controls (added in #104) are not showing up. I'm pretty sure this is related to the duplicate route name issue I already noted above. If you sort that out, I think you'll start to see more information on all the object editor pages.

3.) I was going to test this by uploading a thumbnail image to a newly created empty folder... but when I create a folder, the only datastream controls I get are RELS-EXT and DC. I haven't had time to dig into the front-end code yet to figure out why this might be... but based on the fact that every object includes CoreModel, and CoreModel allows THUMBNAIL datastreams (and your configuration seems to reflect this), I'm not sure why the option is missing.

api/src/services/Datastream.test.ts
diboy2: made change

api/src/services/Datastream.ts
diboy2: Made change

api/src/services/FedoraCatalog.ts
diboy2: So, per allowed type there are multiple subtypes. That's why I kept it singular, like "image/png" or "image/jpeg".

api/src/services/HierarchyCollector.ts
diboy2: made change

api/vudl.ini.dist

diboy2:
@demiankatz I made the assumption that I was supposed to use the datastreams from HierarchyCollector to display the datastream controls, but it seems it's supposed to utilize catalog. I'm going to make a change to use the model's datastreams in the catalog.

api/src/services/FedoraCatalog.ts
demiankatz: True. I guess I was just thinking of the case where allowedType is a wildcard, in which case multiple values will match... but until there's a use case for a comma-separated list of types (which so far there is not), this is probably fine. :-)

api/src/routes/edit.ts
diboy2: Made change to /modelsdatastreams/

demiankatz: Thanks! In the future, we may want to look into whether we should combine some of these things to reduce the number of API calls, but for now this makes sense. Of course I realize these are the sorts of issues that GraphQL can address, though best to keep it simple for now.

api/src/services/Datastream.ts
```diff
@@ -0,0 +1,39 @@
+import { FedoraObject } from "../models/FedoraObject";
+import FedoraCatalog from "./FedoraCatalog";
+
+class Datastream {
```
demiankatz: Nitpick, but would DatastreamManager be a better name, since this is a service that manages datastreams rather than a model of a datastream?

api/src/services/Datastream.ts
```diff
@@ -0,0 +1,39 @@
...
+        if (this.hasValidMimeType(stream, mimeType)) {
+            await fedoraObject.updateDatastreamFromFile(filepath, stream, mimeType);
+        } else {
+            throw new Error(`Invalid mime type: ${mimeType}`);
+        }
+    }
+
+    hasValidMimeType(stream: string, mimeType: string): boolean {
+        const [type, subtype] = mimeType.split("/");
+        const { allowedType, allowedSubtypes } = this.fedoraCatalog.getDatastreamMimetypes()?.[stream]?.mimetype;
```
demiankatz: Does this behave gracefully if everything is undefined? Seems like allowedType/allowedSubtypes might end up being null, and then some of the code below might fail. Is there a test to cover that scenario?

api/vudl.ini.dist
```diff
@@ -194,3 +194,93 @@ DOC = "vudl-system:DOCData"
 Audio = "vudl-system:AudioData"
 XLS = "vudl-system:XLSData"
 Text = "vudl-system:TextData"
+
+# AudioData
+[models.AudioData.datastreams.WAV.mimetype]
+allowedType = "audio"
+allowedSubtypes = "wav,x-wav"
+
+
```
demiankatz: There seem to be a lot of newlines in the additions to this file. Might be good to remove some to clean it up a bit. I'd suggest getting rid of any double-breaks, and being consistent about whether or not there is a blank line after the header comments like `# AudioData`.

client/components/edit/ObjectEditor.tsx
```diff
@@ -1,20 +1,42 @@
...
+
+    useEffect(() => {
+        setCurrentPid(pid);
+        initializeModelsCatalog();
+    }, []);
+
     return (
         <div>
             <h1>Editor: Object {pid}</h1>
-            <ObjectSummary pid={pid} />
```
demiankatz, line 11: Did you intentionally remove the Object Summary, or was this a merge error?

client/components/edit/datastream/DatastreamModal.tsx
```diff
@@ -0,0 +1,32 @@
+import React from "react";
+import { Dialog, DialogTitle, Grid, IconButton } from "@mui/material";
+import CloseIcon from "@mui/icons-material/Close";
+import { useEditorContext } from "../../../context/EditorContext";
+import DatastreamUploadModalContent from "./DatastreamUploadModalContent";
+
+const DatastreamModal = () => {
```
demiankatz, line 7: Is this a DatastreamUploadModal rather than just a DatastreamModal, or do you plan to embed all datastream functionality into this one component in future?

client/components/edit/datastream/DatastreamUploadModalContent.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+                    ],
+                },
+            });
+            wrapper.update();
+        });
+
+        expect(wrapper.children().text()).toContain("Illegal mime type");
+    });
+
+    it("renders illegal mime type when catalog cannot fin datastream", async () => {
```
demiankatz: Typo -- "fin" should be "find".

client/context/EditorContext.tsx
```diff
@@ -0,0 +1,193 @@
...
+    currentModels: [],
+    currentDatastreams: [],
+    currentDatastream: null,
+    isDatastreamModalOpen: false,
+    datastreamModalState: null,
+};
+
+export const DatastreamModalStates = {
+    UPLOAD: "Upload",
+    VIEW: "View",
```
demiankatz: Should you comment out all the states other than UPLOAD for now, and then uncomment them when they're implemented? It's a bit confusing that all the buttons appear but they all just upload for now. Might be worth hiding them until they're ready, if we want to do an incremental merge for just upload functionality.

client/context/EditorContext.tsx
```diff
@@ -0,0 +1,193 @@
...
+/**
+ * Pass a shared entity to react components,
+ * specifically a way to make api requests.
+ */
+const editorContextParams = {
+    modelsCatalog: {},
+    currentPid: null,
+    currentModels: [],
+    currentDatastreams: [],
+    currentDatastream: null,
```
demiankatz: Having variables called currentDatastream and currentDatastreams seems potentially confusing. If currentDatastreams is all the datastreams present on the current object, and currentDatastream is the one the user is interacting with presently, maybe renaming currentDatastream to activeDatastream would better differentiate the naming.

demiankatz (changes requested):
Thanks, @diboy2! I had time to review this in a bit more detail, and I left a few more comments. As usual, all minor points. :-)

client/components/edit/ObjectEditor.tsx
diboy2: I think so. I'll make the changes.

client/components/edit/datastream/DatastreamModal.tsx
diboy2: I'm planning provide a mapping of this later. It's in my next pr.

client/context/EditorContext.tsx
diboy2: I'll make that change.

api/src/services/Datastream.ts
diboy2: Made CHange.

api/vudl.ini.dist
diboy2: Made chnage.

diboy2: Pooled all the same models together.

client/components/edit/datastream/DatastreamUploadModalContent.test.tsx
diboy2: Made change.

client/context/EditorContext.tsx
diboy2: Made change.

api/src/services/Datastream.ts
diboy2: Refactored with optional chaining safeguards. Added unit tests.

demiankatz (approved):
Thanks for the latest fixes and adjustments -- this is looking good to me now; I believe it is ready to merge.

Speaking of merging, did you want to take another look at #92 (the PR to add a breadcrumb trail as you navigate down the folder tree)? I believe it is ready to merge as well, though I imagine we'll have to resolve some conflicts between it and this.

PR #112, opened 2022-03-03
Fix problem with bullmq tests adding real data to Redis.

demiankatz:


demiankatz:
@diboy2, working on tests reminded me that we still have an open issue where the test suite causes real data to be added to Redis via bullmq. This PR is an attempt to make the mocking work right. Not sure if this is the best approach, but it works in the sense that tests still pass and no data is sent to Redis. Could you review when time permits?

api/src/services/QueueManager.test.ts
```diff
@@ -1,7 +1,14 @@
 import QueueManager from "./QueueManager";
 import { Queue } from "bullmq";
 
-jest.mock("bullmq");
+jest.mock("bullmq", () => {
+    return {
+        Queue: jest.fn(),
```
diboy2, line 6: can you try? 

`
Queue : { 
 add: jest.fn(),
 close: jest.fn(),
 getJobs: jest.fn()
}
`

demiankatz: Unfortunately, that's one of the first things I tried, but it doesn't work. I end up with this error:

<pre>
    Cannot spyOn on a primitive value; undefined given

      20 |         beforeEach(() => {
      21 |             jobs = [];
    > 22 |             getJobsSpy = jest.spyOn(Queue.prototype, "getJobs").mockResolvedValue(jobs);
         |                               ^
      23 |             addSpy = jest.spyOn(Queue.prototype, "add");
      24 |         });
      25 |

      at ModuleMocker.spyOn (node_modules/jest-mock/build/index.js:785:13)
      at Object.<anonymous> (src/services/QueueManager.test.ts:22:31)
</pre>

...so then my next thought was to define the prototype like:

<pre>
jest.mock("bullmq", () => {
    return {
        Queue: { prototype: { add: jest.fn(), close: jest.fn(), getJobs: jest.fn() } }
    };
});
</pre>

...but then that fails with:

<pre>
    TypeError: bullmq_1.Queue is not a constructor

      53 |
      54 |     public async queueMetadataOperation(pid: string, action: string): Promise<void> {
    > 55 |         const q = new Queue(this.defaultQueueName);
         |                   ^
      56 |         const jobs = await q.getJobs("wait");
      57 |         const queueJob = { pid, action };
      58 |         if (this.isAlreadyAwaitingAction(jobs, "metadata", queueJob)) {

      at QueueManager.<anonymous> (src/services/QueueManager.ts:55:19)
      at step (src/services/QueueManager.ts:4718:19)
      at Object.next (src/services/QueueManager.ts:4448:14)
      at src/services/QueueManager.ts:4376:40
      at Object.<anonymous>.__awaiter (src/services/QueueManager.ts:4319:10)
      at QueueManager.queueMetadataOperation (src/services/QueueManager.ts:5166:12)
      at src/services/QueueManager.test.ts:35:32
      at step (src/services/QueueManager.test.ts:33:23)
      at Object.next (src/services/QueueManager.test.ts:14:53)
</pre>

...and that's how I ended up with the syntax used in this PR. I don't like it, but it's the only combination I could find that actually worked as expected.

diboy2: ok

diboy2 (approved):
👍

demiankatz:
Thanks, @diboy2. I've gone ahead and merged as-is; we can refactor later if we find a better way to represent this.

PR #113, opened 2022-03-08
Fix useAutoLayoutEffect on jest

diboy2:


demiankatz (approved):

demiankatz:
Thanks, @diboy2 -- looks fine to me, so feel free to merge at will.

One other thought that occurred to me while reviewing the CI logs: I wonder if we should mock console and assert about messages on tests that have console output, to both test that messages are sent as expected and also to declutter the actual output of the tests. Not urgent, but maybe something to think about going forward.

crhallberg (approved):

PR #114, opened 2022-03-10
Make VuFind URL optional for indexing.

demiankatz:
This should make it easier for @diboy2 (or anyone else) to test the system without having to either fully configure VuFind or else comment out VuFind-specific code in the indexer. Now we can just unset the setting and it will omit unsupported data gracefully.


diboy2 (approved):
Thanks!!! 👍

PR #115, opened 2022-03-11
Add delete datastream feature

diboy2:


api/src/models/FedoraObject.ts
```diff
@@ -53,6 +53,15 @@ export class FedoraObject {
         await this.fedora.addDatastream(this.pid, id, params, data, expectedStatus);
     }
 
+    async deleteDatastream(stream: string): Promise<void> {
+        await this.fedora.deleteDatastream(this.pid, stream);
+        await this.fedora.deleteDatastreamTombstone(this.pid, stream);
```
demiankatz: Perhaps apples and oranges, but would it be better to just call `this.deleteDatastreamTombstone(stream)` here to reduce redundancy?

api/src/routes/sanitize.ts
```diff
@@ -15,4 +15,4 @@ export function sanitizeParameters(customRules = {}, defaultRule = /^[-.a-zA-Z0-
     };
 }
 
-export const pidSanitizer = sanitizeParameters({ pid: /^[a-zA-Z]+:[0-9]+/ }, /^$/);
+export const pidSanitizer = sanitizeParameters({ pid: /^[a-zA-Z]+:[0-9]+/ });
```
demiankatz: I wonder if we need to rethink the sanitizeParameters function to make it easier to chain things. For example, maybe if we pass null or false to defaultRule, it entirely bypasses checking for parameters outside of customRules. That would allow us to use the defined pidSanitizer to sanitize PIDs, but then chain a new purpose-built function created by sanitizeParameters to handle any additional parameters.

There are really a lot of different ways I could think of to approach this, and I'm not sure what's best... but I want to be sure we don't paint ourselves into a corner.

api/src/services/DatastreamManager.ts
```diff
@@ -18,13 +18,27 @@ class DatastreamManager {
...
     async uploadFile(pid: string, stream: string, filepath: string, mimeType: string): Promise<void> {
         const fedoraObject = FedoraObject.build(pid);
+
         if (this.hasValidMimeType(stream, mimeType)) {
-            await fedoraObject.updateDatastreamFromFile(filepath, stream, mimeType);
+            try {
+                await fedoraObject.updateDatastreamFromFile(filepath, stream, mimeType);
+            } catch (error) {
+                console.log(error);
+                if ((error?.message || "").includes("410")) {
```
demiankatz: I think this is a reasonably safe test, but I wonder if we could make it more explicit by using a more specific exception class for HTTP errors and storing the HTTP code in a property on the thrown exception object.

demiankatz (commented):
I plan to take a deeper look at this next week, but here are a few initial thoughts (focused on the API side of the code).

client/util/routes.ts
```diff
@@ -37,6 +37,10 @@ const postObjectDatastreamUrl = (pid) => {
     return `${editObjectUrl}/${pid}`;
 }
 
+const deleteObjectDatastreamUrl = (pid, datastream: string) => {
+    return `${editObjectUrl}/${pid}/datastream/${datastream.toLowerCase()}`;
```
demiankatz: Why are you lower-casing the datastream name here and then uppercasing it again in the route? While in practice all datastream names are currently ALLCAPS, I think it might be safer to assume that they are case-sensitive for future flexibility.

client/components/edit/datastream/DatastreamControlButton.tsx
```diff
@@ -15,14 +24,17 @@ const DatastreamControlButton = ({ modalState, datastream }) => {
     };
 
     return (
-        <Button onClick={onClick} variant="outlined" size="small">
-            {modalState}
-        </Button>
+        <Tooltip title={modalState}>
+            <IconButton disabled={modalState !== "Upload" && disabled} onClick={onClick} size="small">
+                {Icons[modalState]}
```
demiankatz, line 34: Should we put some alt-text or tool tips on these?

demiankatz (commented):
I gave this another look this morning, focusing more heavily on the React code, and I had a couple more small comments. Looking great overall, though!

client/util/routes.ts
diboy2: I think it was a stylistic choice since I don't usually see CAPS LOCK characters in routes. But, after investigating how CAPS LOCK characters in urls would affect the app, I'm just going to remove the lowercase and uppercase operations.

api/src/models/FedoraObject.ts
diboy2: Oh yeah. lol

api/src/services/DatastreamManager.ts
diboy2: Made change at the Fedora.ts level.

client/components/edit/datastream/DatastreamControlButton.tsx
diboy2: Added aria-label for alternative text for <svg> icons.

api/src/routes/sanitize.ts
diboy2: I removed the regex "^$" because it was disallowing all values from additional params. There many ways to implement this. Atm, I can't think up those options. I think it requires further investigation and maybe should be delegated to another task.

demiankatz: How do you feel about b7417b3398603a4a4f4593e34fce64aa1bc192a9? This gets the behavior we need without changing as much existing behavior. There's still probably room to refactor/simplify, but I think this makes the current PR a little bit less impactful, and we can optimize the rest later.

demiankatz (approved):

demiankatz:
Thanks, @diboy2, this is looking great! Just a few small comments:

1.) I fixed the Solr indexing bug revealed by this work in [e73a33b](https://github.com/FalveyLibraryTechnology/VuDLPrepJS/pull/115/commits/e73a33b67525791a68836bf5187a79d305c414c5)

2.) There was a broken test related to the case-sensitive datastream change. I fixed that in [16e4529](https://github.com/FalveyLibraryTechnology/VuDLPrepJS/pull/115/commits/16e452968e9828d9fefabfa1e298c1c9eb0b128c).

I believe this is now ready to merge!

api/src/routes/sanitize.ts
diboy2: Sure. Let's also create a followup story.

demiankatz: I've created a task for us in Teams. We may at some point want to begin moving some things from Teams to GitHub issues if it makes more sense to discuss them in a more public setting. At this early stage, though, I don't think it makes a difference yet.

PR #116, opened 2022-03-17
Handle Follow messages from Camel Toolbox.

demiankatz:
I've just encountered a new type of message coming from the Camel Toolbox, caused by the creation of certain types of relationships. (When resource B is assigned certain triples that refer to resource A, then resource A receives a Follow event). I think this is happening as a result of changes in #98, since we are expressing more object relationships in RDF instead of the RELS-EXT datastream. In any case, I do not believe that we need to trigger a reindex in these situations, so this PR simply ignores the messages to avoid filling the console with warnings.

### DIFF
--- a/api/src/routes/messenger.ts
+++ b/api/src/routes/messenger.ts
@@ -70,6 +70,9 @@ messenger.post("/camel", bodyParser.json(), async function (req, res) {
         case "Purge":
             await QueueManager.getInstance().performIndexOperation(pid, "delete");
             break;
+        case "Follow":
+            // Follow messages do not require any action at this time.
+            break;
         default: {
             const msg = "Unexpected action: " + action + " (on PID: " + pid + ")";
             console.error(msg);


demiankatz:
Requesting a review from @diboy2 and @crhallberg, but I realize this change is pretty deep in the rabbit hole. Happy to discuss if you have any questions.

api/src/routes/messenger.ts
```diff
@@ -70,6 +70,9 @@ messenger.post("/camel", bodyParser.json(), async function (req, res) {
         case "Purge":
             await QueueManager.getInstance().performIndexOperation(pid, "delete");
             break;
+        case "Follow":
```
diboy2, line 73: I don't think you need this additional break statement. You can just move it one line up before the previous break statement.

demiankatz: That's true, though there may be a stylistic advantage to keeping it separated like this (e.g. if we eventually decide to add functionality to this case, we don't want the previous case to drop through into it... and that might be an easy mistake to make otherwise).

diboy2: ok

diboy2 (approved):

demiankatz:
Thanks, @diboy2!

PR #117, opened 2022-03-21
Create test coverage for substring calls

demiankatz:
This PR is a companion to #106, intended to improve test coverage of methods calling substr in preparation for replacing the code with a non-deprecated method.

### DIFF
--- a/api/src/jobs/Ingest.test.ts
+++ b/api/src/jobs/Ingest.test.ts
@@ -61,29 +61,22 @@ describe("IngestProcessor", () => {
     });
     describe("moveDirectory", () => {
         it("moves the directory appropriately", () => {
-            const date = new Date();
-            const month = "0" + (date.getMonth() + 1);
-            const day = "0" + date.getDate();
-            const expectedTargetParent =
-                "/fake_processed/" +
-                date.getFullYear() +
-                "-" +
-                month.substring(month.length - 2) +
-                "-" +
-                day.substring(day.length - 2) +
-                "/fake";
+            const expectedTargetParent = "/fake_processed/2\\d\\d\\d-\\d\\d-\\d\\d/fake";
+            const expectedTargetParentRegEx = new RegExp(expectedTargetParent);
             const expectedTarget = expectedTargetParent + "/fakejob";
+            const expectedTargetRegEx = new RegExp(expectedTarget);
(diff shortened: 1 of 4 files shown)


demiankatz:
Note that some of the tests added in #98 are also necessary to cover all of #106; I am not going to duplicate them here.

demiankatz:
@diboy2, this PR doesn't fully cover all of the changes in #106, but it does improve coverage overall. If you're happy with it, I think we should merge it. I've put more specific TODO checkboxes on #106 to highlight the areas where testing is still needed (the HierarchyCollector and SolrIndexer, both of which are complex to test and probably need some refactoring to improve testability -- I'll likely tackle those things in separate PRs to keep things manageable).

api/src/jobs/Ingest.test.ts
```diff
@@ -0,0 +1,89 @@
...
+            expect(replaceSpy).toHaveBeenCalledWith(
+                fedoraObject,
+                "<oai:dc><dc:title>fakejob_dir_fake_my</dc:title></oai:dc>",
+                "Set dc:title to ingest/process path"
+            );
+        });
+    });
+    describe("moveDirectory", () => {
+        it("moves the directory appropriately", () => {
+            const date = new Date();
```
diboy2: Are you trying to use the current date whenever this test runs? Is it possible to use fake timers to run the unit test?

demiankatz: I think a better solution is actually to use a pattern matcher on the expectation for the filename, because I want to be sure there's a date in the resulting output, but I don't really care what date it is. I wasn't able to figure out how to do that before, so I did this out of laziness, but I think I have an idea of how to improve it... I'll try momentarily.

demiankatz: Okay, see [cb83ec5](https://github.com/FalveyLibraryTechnology/VuDLPrepJS/pull/117/commits/cb83ec5314b1bcbb12a2613be01387cbc452d802) -- I think that's much nicer. (I could probably make my regular expression more concise, but I had trouble getting the code to recognize the `{n}` syntax for repeating patterns, and rather than waste a lot of time troubleshooting, I just made a somewhat repetitive regex).

diboy2 (approved):

PR #118, opened 2022-03-23
Add download datastream feature

diboy2:


api/src/models/FedoraObject.test.ts
```diff
@@ -36,6 +36,18 @@ describe("FedoraObject", () => {
         });
     });
 
+    describe("getDatastreamMetadata", () => {
+        let getRdfSpy;
+        it("gets the datastream metadata", async () => {
+            getRdfSpy = jest.spyOn(Fedora.prototype, "getRdf").mockResolvedValue("test");
```
demiankatz, line 82: Not necessarily a request for changes, but just a point to discuss best practice on: when working on tests for #98, I ran into some problems caused by mocks bleeding from one test into another. I solved it by adding an `afterEach` to call `jest.clearAllMocks();`. I wonder if that's a good general habit to form in tests, or if it's something we should only do when we absolutely have to.

api/src/models/FedoraObject.test.ts
```diff
@@ -36,6 +36,18 @@ describe("FedoraObject", () => {
...
+
+            const metadata = await fedoraObject.getDatastreamMetadata(stream);
+
+            expect(metadata).toEqual("test");
+            expect(getRdfSpy).toHaveBeenCalledWith(expect.stringContaining(stream));
```
demiankatz: Is it worth making a more specific assertion here? Any reason NOT to match the full URL being passed in?

api/src/routes/edit.test.ts
```diff
@@ -46,4 +46,29 @@ describe("edit", () => {
             expect(datastreamManager.deleteDatastream).toHaveBeenCalledWith(pid, datastream);
         });
     });
+
+    xdescribe("get /object/:pid/datastream/:stream/download", () => {
```
demiankatz: Is this test intentionally disabled with `xdescribe`?

api/src/routes/edit.test.ts
```diff
@@ -46,4 +46,29 @@ describe("edit", () => {
...
+        let datastreamManager;
+        beforeEach(() => {
+            datastreamManager = {
+                getMimeType: jest.fn(),
+                downloadBuffer: jest.fn(),
+            };
+            jest.spyOn(Database, "confirmToken").mockResolvedValue(true);
+            jest.spyOn(DatastreamManager, "getInstance").mockReturnValue(datastreamManager);
```
demiankatz, line 136: When writing my own tests, I got type errors when I tried to substitute fake objects for classes using this pattern. I'm not sure how you're able to get away with this. Maybe I should just have ignored the warnings in my IDE... or maybe we should do some kind of typecasting to make Typescript happier about it. (Or maybe I was making a mistake in my own code, and there is no problem at all here).

api/src/routes/edit.test.ts
```diff
@@ -46,4 +46,29 @@ describe("edit", () => {
...
+        });
+
+        it("will download the datastream content", async () => {
+            datastreamManager.getMimeType.mockResolvedValue("test1");
+            datastreamManager.downloadBuffer.mockResolvedValue("test2");
+
+            await request(app)
+                .get(`/edit/object/${pid}/datastream/${datastream}/download`)
+                .set("Authorization", "Bearer test")
+                .expect(StatusCodes.OK);
```
demiankatz, line 148: Should we actually capture the response in a variable and make some assertions about its content? I would expect we could at least double-check appropriate content-type header and body contents.

api/src/routes/edit.ts
```diff
@@ -154,6 +154,25 @@ edit.get("/object/parents/:pid", pidSanitizer, requireToken, async function (req
     }
 });
 
+edit.get("/object/:pid/datastream/:stream/download", pidSanitizer, requireToken, async function (req, res) {
```
demiankatz: Shouldn't you be using datastreamSanitizer rather than pidSanitizer? I wouldn't expect this to work as written because pidSanitizer should block the stream parameter.

api/src/routes/edit.ts
```diff
@@ -154,6 +154,25 @@ edit.get("/object/parents/:pid", pidSanitizer, requireToken, async function (req
...
+    const pid = req.params.pid;
+    const stream = req.params.stream;
+    const datastream = DatastreamManager.getInstance();
+    try {
+        const fileType = await datastream.getMimeType(pid, stream);
+        const buffer = await datastream.downloadBuffer(pid, stream);
+
+        res.writeHead(200, {
+            "Content-Type": fileType,
+            "Content-Length": buffer.length,
```
demiankatz: Do we need to write the Content-Length, or can we trust the framework to do it for us? (I'm not sure of the answer, but just wondering if a simplification is possible).

I'm also curious whether we should add a [Content-Disposition](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition) header to ensure that things get downloaded rather than displayed inline. (Conversely, it might also be useful to have a parameter on this route which could be used to switch to inline -- then we could recycle the same code if we want to display a datastream inline later).

api/src/services/DatastreamManager.test.ts
```diff
@@ -75,6 +76,56 @@ describe("DatastreamManager", () => {
...
+            pid = "test1";
+            stream = "test2";
+            xml = "test3";
+            getDatastreamMetadataSpy = jest.spyOn(FedoraObject.prototype, "getDatastreamMetadata");
+        });
+
+        it("gets the datastream metadata", async () => {
+            getDatastreamMetadataSpy.mockResolvedValue(xml);
+            const metadata = await datastreamManager.getMetadata(pid, stream);
+            expect(metadata).toEqual(xml);
```
demiankatz, line 95: Should we also expect getDatastreamMetadataSpy to have been called once with particular parameters?

api/src/services/DatastreamManager.ts
```diff
@@ -16,6 +17,24 @@ class DatastreamManager {
...
+    }
+
+    async getMimeType(pid: string, stream: string): Promise<string> {
+        const metadataExtractor = MetadataExtractor.getInstance();
+        const xml = await this.getMetadata(pid, stream);
+        const ebuNode = metadataExtractor.extractEbuNode(xml, "//ebucore:hasMimeType");
+        return ebuNode?.hasMimeType?.[0] || "";
+    }
+
+    async downloadBuffer(pid: string, stream: string): Promise<Buffer> {
```
demiankatz, line 33: Is test coverage missing for this new method, or did I just overlook something?

api/src/services/MetadataExtractor.test.ts
```diff
@@ -0,0 +1,34 @@
+import MetadataExtractor from "./MetadataExtractor";
+import { DOMParser } from "@xmldom/xmldom";
+
+describe("MetadataExtractor", () => {
```
demiankatz, line 4: I also created a MetadataExtractor.test.ts in #98; we'll have to reconcile them at some point (though I don't expect they'll cause problems for each other).

api/src/services/MetadataExtractor.test.ts
```diff
@@ -0,0 +1,34 @@
...
+            parseFromStringSpy = jest.spyOn(DOMParser.prototype, "parseFromString");
+        });
+
+        it("extracts an ebu node", async () => {
+            parseFromStringSpy.mockReturnValue(rdfXml);
+            const extractRDFXMLSpy = jest.spyOn(metadataExtractor, "extractRDFXML").mockReturnValue({
+                test: ["test4"],
+            });
+
+            metadataExtractor.extractEbuNode(xml, xpathQuery);
```
demiankatz: Should we check the return value here?

api/src/services/MetadataExtractor.ts
```diff
@@ -220,6 +218,17 @@ class MetadataExtractor {
             "//premis:*"
         );
     }
+
+    public extractEbuNode(xml: string, xpathQuery: string): Record<string, Array<string>> {
```
demiankatz: Nitpick, but I wonder if there's a better name for this method -- perhaps `extractEbuNodes` (since there could be more than one) or `extractEbuCore` (to reflect the full name of the ontology being used, which might be more consistent with other existing methods).

client/context/FetchContext.tsx
```diff
@@ -132,11 +132,24 @@ export const useFetchContext = () => {
         throw new Error(response.statusText);
     };
 
+    const fetchBlob = async (url = ingestApiUrl, params = {}, headers = {}) => {
```
demiankatz: Does it really make sense to have a default value for `url` here? Would it be better for that to be a mandatory parameter?

client/hooks/useDatastreamOperation.test.ts
```diff
@@ -0,0 +1,220 @@
...
+                writable: true,
+            });
+            Object.defineProperty(global, "URL", {
+                value: {
+                    createObjectURL: createObjectURLSpy,
+                },
+                writable: true,
+            });
+        });
+        xit("success", async () => {
```
demiankatz: Is this intentionally disabled?

client/hooks/useDatastreamOperation.test.ts
```diff
@@ -0,0 +1,220 @@
...
+            headers.append("Content-Type", "image/jpeg");
+            fetchValues.action.fetchBlob.mockResolvedValue({
+                headers,
+                blob
+            });
+            const { downloadDatastream } = useDatastreamOperation();
+            await downloadDatastream("test1");
+
+            // expect(global.document.createElement).toHaveBeenCalled();
```
demiankatz: Should this be uncommented or removed?

client/package.json
```diff
@@ -15,6 +15,7 @@
     "@fortawesome/fontawesome-free": "^5.15.4",
     "@material-ui/lab": "^4.0.0-alpha.60",
     "@mui/icons-material": "^5.4.1",
+    "@mui/lab": "^5.0.0-alpha.73",
```
demiankatz, line 17: What is the difference between `@material-ui/lab` and `@mui/lab`? Should we be using one or the other rather than both, or are they truly separate things?

client/hooks/useDatastreamOperation.ts
```diff
@@ -0,0 +1,104 @@
...
+    };
+
+    const downloadDatastream = async (datastream) => {
+        try {
+            const response = await fetchBlob(downloadObjectDatastreamUrl(currentPid, datastream));
+            const mimeType = response?.headers?.get("Content-Type")?.split("/");
+            if(mimeType?.[1] || response?.blob) {
+                throw new Error("Incorrect file format");
+            }
+            const link = document.createElement("a");
```
demiankatz, line 83: If you have appropriate Content-Disposition headers set by the route, I wonder if you could just do a `Window.open` on the blob URL rather than having to create and click on a link. Might allow for some further simplification -- among other things, you wouldn't have to worry about pre-fetching the mime type.

demiankatz (changes requested):
Thanks, @diboy2! I haven't done hands-on tests yet, but here are some comments based on an initial code read. As always, it looks good, and I like the refactoring you've done, but I have a few comments and suggestions.

api/src/routes/edit.test.ts
diboy2: Oh. I think I missed this.

client/hooks/useDatastreamOperation.ts
diboy2: I think I chose it this way, since I was still able to pass the token using this method.

client/package.json
diboy2: Materia-ui lab is the depecrated version. We should be using the @mui/lab version.

diboy2: I'll make the change to move all references.

api/src/models/FedoraObject.test.ts
diboy2: I think it's good practice to add clearAllMocks to test suites. But, maybe we shouldn't rely too much on mocking prototypes and use more dependency injection to mock our entities.

api/src/models/FedoraObject.test.ts
diboy2: I think I was just being lazy her.

api/src/routes/edit.test.ts
diboy2: I think I did a mock DatastreamManager first in the beginning of this file.

api/src/routes/edit.test.ts
diboy2: Making change.

client/context/FetchContext.tsx
diboy2: Making change.

client/hooks/useDatastreamOperation.test.ts
diboy2: Making change.

client/hooks/useDatastreamOperation.ts
demiankatz: I see! That is probably enough reason to leave it alone -- but if we run into trouble with this approach, the other option might be worth revisiting in the future.

client/package.json
demiankatz: Thanks! I think we're only using the old version in one place: for the tree-based model selection control in the "create object" component.

api/src/routes/edit.test.ts
demiankatz: Yeah, that's probably the difference. Still trying to get a handle on how to do mocks in a type-appropriate way!

api/src/models/FedoraObject.test.ts
demiankatz: I agree that dependency injection is preferable when possible... though again, I sometimes run into type problems; I'm not sure how to create a mock instance of an object without mocking the whole module the class comes from. I've generally had to create "real" objects first and then mock out the methods I want to control. (But maybe that's effectively the same thing... it just would be nicer if I could say in a simple way "give me a mock version of class X with all the methods but none of the functionality" and I haven't found that).

diboy2: Made change.

api/src/models/FedoraObject.test.ts
diboy2: Made change.

api/src/routes/edit.ts
diboy2: Made change.

api/src/services/DatastreamManager.test.ts
diboy2: Made change.

api/src/services/DatastreamManager.ts
diboy2: Made change.

api/src/services/MetadataExtractor.test.ts
diboy2: Made change.

api/src/services/MetadataExtractor.ts
diboy2: Changed to extractEbuCore.

client/package.json
diboy2: Made change.

api/src/routes/edit.ts
diboy2: I removed content length and used res. send. If I use the Content-disposition approach, I have to store the data into a file first and pass the file name. I went for this approach so I could pass the token and not have to store the file.

client/hooks/useDatastreamOperation.test.ts
diboy2: Made change

api/src/routes/edit.test.ts
```diff
@@ -58,15 +62,17 @@ describe("edit", () => {
...
+            datastreamManager.getMimeType.mockResolvedValue(fileType);
+            datastreamManager.downloadBuffer.mockResolvedValue(buffer);
 
-            await request(app)
+            const response = await request(app)
                 .get(`/edit/object/${pid}/datastream/${datastream}/download`)
                 .set("Authorization", "Bearer test")
                 .expect(StatusCodes.OK);
 
+            //  expect(response.headers["Content-Type"]).toEqual(fileType);
```
demiankatz: Is there a reason this is commented out?

api/src/routes/edit.test.ts
```diff
@@ -58,15 +62,17 @@ describe("edit", () => {
             jest.spyOn(DatastreamManager, "getInstance").mockReturnValue(datastreamManager);
         });
 
-        it("will download the datastream content", async () => {
-            datastreamManager.getMimeType.mockResolvedValue("test1");
-            datastreamManager.downloadBuffer.mockResolvedValue("test2");
+        xit("will download the datastream content", async () => {
```
demiankatz: Is this intentionally disabled?

client/context/FetchContext.tsx
demiankatz: Looks like this change did not get made (or at least, not pushed up yet).

demiankatz (changes requested):
@diboy2, thanks for all the progress on this. I've resolved all the open conversations that have been taken care of, but I spotted a couple of things that still seem to be in progress. (Apologies for repeating myself if you're already in the process of working on these).

I tried a bit of hands-on testing but was unable to get downloads to work. I think this may be due to the fact that my test environment is now set up for the changes in #98, and I did a sloppy merge to try to test this code. If you have time to finish up the review of #98 (since I replied to the comments you left), it would be really helpful if we could get that merged, and then I can try testing again without having to rebuild my test Fedora instance.

api/src/routes/edit.test.ts
diboy2: weird. I thought I changed that.

diboy2: Made change.

api/src/routes/edit.test.ts
diboy2: Made chnage.

client/hooks/useDatastreamOperation.ts
```diff
@@ -0,0 +1,104 @@
...
+                severity: "error",
+            });
+        }
+        toggleDatastreamModal();
+    };
+
+    const downloadDatastream = async (datastream) => {
+        try {
+            const response = await fetchBlob(downloadObjectDatastreamUrl(currentPid, datastream));
+            const mimeType = response?.headers?.get("Content-Type")?.split("/");
```
demiankatz: This needs to be made a little smarter, to handle MIME types like `text/xml; charset=utf8`.

demiankatz: See [73236a5](https://github.com/FalveyLibraryTechnology/VuDLPrepJS/pull/118/commits/73236a586e6355e49c35fbdaac70a832125192c9) -- I believe this addresses the problem. If you're happy with this, I believe the PR is ready to merge.

demiankatz (approved):

api/src/routes/edit.ts
```diff
@@ -141,6 +141,7 @@ edit.get("/object/details/:pid", requireToken, pidSanitizer, async function (req
     const extractedMetadata = MetadataExtractor.getInstance().extractMetadata(metadata);
     res.json({ pid, sort, metadata: extractedMetadata });
 });
+const datastreamSanitizer = sanitizeParameters({ pid: pidSanitizeRegEx, stream: defaultSanitizeRegEx }, /^$/);
 
 edit.get("/object/parents/:pid", pidSanitizer, requireToken, async function (req, res) {
```
crhallberg: 1. Would it make sense to use the datasteamSanitizer in this case, to reduce redundancy?
2. Should we move the datastreamSanitizer into `sanitize.ts` where the pidSanitizer and these regexs are created?

crhallberg (commented):
Good after my one comment is addressed.

api/src/routes/edit.ts
demiankatz: I'm not sure I understand your first question -- where are you suggesting we should use the datastreamSanitizer where it is not already being used?

Regarding your second point, that probably makes sense -- we have an open task on our project board to rework/improve our sanitization strategy, but that's a small step that probably makes this slightly cleaner.

api/src/routes/edit.ts
```diff
@@ -152,7 +153,27 @@ edit.get("/object/parents/:pid", pidSanitizer, requireToken, async function (req
...
 
-const datastreamSanitizer = sanitizeParameters({ pid: pidSanitizeRegEx, stream: defaultSanitizeRegEx }, /^$/);
+edit.get("/object/:pid/datastream/:stream/download", datastreamSanitizer, requireToken, async function (req, res) {
+    const pid = req.params.pid;
+    const stream = req.params.stream;
+    const datastream = DatastreamManager.getInstance();
+    try {
+        const mimeType = await datastream.getMimeType(pid, stream);
+        const fileType = mimeType?.split("/")?.[1];
+        const fileName = `${pid}_${stream}.${fileType}`;
```
demiankatz: I'd suggest running a regex on PID to replace non-word characters with underscores, since having colons in filenames may cause problems on certain operating systems.

api/src/routes/edit.ts
crhallberg: Here we are using `pidSanitizer`, which is the same regex and key as half of `datastreamSanitizer`. I guess having two doesn't make additional work, as long as they continue to be based on the same regex pattern.

demiankatz: The difference between pidSanitizer and datastreamSanitizer is that pidSanitizer blocks all parameters other than pid, whereas datastreamSanitizer accepts pid + datastream. If we want to keep the validation as strict as possible, I don't think we should change it.

crhallberg: Didn't parse the fact that it blocks other parameters. No problem!

crhallberg (approved):
Moving datastreamSanitizer into the same file as pidSanitizer is the only change I'd recommend.

api/src/routes/edit.ts
diboy2: Made change.

demiankatz (approved):
This looks good to me; I think it's ready to merge.

I do see one strange issue: when I download PDFs, they open in my browser instead of downloading. However, this is difficult to test right now because PDFs slow down the interface due to the issues I'm addressing in #120. I would be inclined to merge this now (since the PDF behavior is strange but not a deal-breaker) and investigate PDFs further after #120 is completed.

PR #120, opened 2022-03-30
Refactor Fedora extra details for fetch-on-demand.

demiankatz:
The hierarchy collector was doing a lot of extra work to populate data for use by the Solr indexer... which was very expensive in cases where the data was not needed. This PR refactors everything so that the data is only retrieved "on demand."

TODO
- [x] Check that all refactored fields still populate the index correctly
- [x] Write tests to cover all modified code
- [x] Consider a better name for the FedoraExtraDetails class -- maybe FedoraOnDemandDetails?


demiankatz:
@diboy2, I think I've managed to get test coverage on all of the changed code. A lot of the test coverage is happening indirectly through the SolrIndexerTest, which is more of an integration test than a unit test... but given time constraints, this seemed the most pragmatic way forward. We can always improve the tests later. I also fixed a few unrelated bugs/type issues that cropped up during test development. Let me know what you think, and if you have any thoughts on naming improvements.

api/src/models/Config.ts
```diff
@@ -2,24 +2,40 @@ import fs = require("fs");
...
+                config = {};
+            }
             // ini returns any, but we can cast it to what we need:
-            Config.instance = new Config(config as Record<string, string>);
+            Config.setInstance(new Config(config as Record<string, string>));
         }
         return Config.instance;
     }
 
+    public static setInstance(config: Config): void {
```
diboy2: Is there a need for this extra function when you're just putting in one line of code?

api/src/models/FedoraExtraDetails.ts
```diff
@@ -0,0 +1,91 @@
...
+    protected fullText: Record<string, Array<string>>;
+
+    constructor(data: FedoraData, fedora: Fedora, extractor: MetadataExtractor, tika: TikaExtractor) {
+        this.data = data;
+        this.fedora = fedora;
+        this.extractor = extractor;
+        this.tika = tika;
+    }
+
+    async getLicense(): Promise<LicenseData> {
```
diboy2: Do you have try catches that surround these calls?

api/src/services/SolrIndexer.ts
```diff
@@ -295,47 +295,55 @@ class SolrIndexer {
...
             const fieldName = "agent." + field + "_txt_mv";
-            fields[fieldName] = fedoraData.agents[field];
+            fields[fieldName] = agents[field];
         }
 
-        if (fedoraData.license !== null) {
-            fields["license.mdRef_str"] = fedoraData.license;
+        const license = await fedoraData.getLicense();
+        if (license !== null) {
+            fields["license.mdRef_str"] = license;
```
diboy2, line 306: I think you can do null coalescing for all the code below. `fields["license.mdRef_str"] = await fedoraData.getLicense() ?? fields["license.mdRef_str"];`

api/src/models/Config.ts
demiankatz: I needed a public way to override the static instance (which is really only needed for testing). I suppose I could just make the static itself public, but it felt better to wrap a method around it in case we want to do extra things in future like outputting warnings to the console, etc.

api/src/services/SolrIndexer.ts
demiankatz: My goal here is to only set `fields["license.mdRef_str"]` to a value if there is a non-null license URL retrieved, and to have the key remain undefined in other situations. Since `fields["license.mdRef_str"]` is not defined prior to the assignment, wouldn't your proposed code result in an error related to accessing an undefined object property and/or result in a key being set to a null value instead of completely omitted? (Javascript is not my primary language, so perhaps it just behaves differently than I assume it behaves -- if that's the case, I apologize).

api/src/models/FedoraExtraDetails.ts
demiankatz: I believe that the Express framework puts an implicit try..catch around everything, and I have put explicit try..catch blocks into route definitions where appropriate (see https://github.com/FalveyLibraryTechnology/VuDLPrepJS/blob/dev/api/src/routes/messenger.ts#L15-L37, which is where this code is exclusively used).

api/src/models/Config.ts
diboy2: ok

api/src/models/FedoraExtraDetails.ts
diboy2: ok

api/src/services/SolrIndexer.ts
diboy2: Oh okay.

diboy2 (approved):

PR #121, opened 2022-03-31
Optimize material-ui imports

diboy2:


client/next.config.js
```diff
@@ -1,4 +1,17 @@
 /** @type {import('next').NextConfig} */
 module.exports = {
+  experimental: {
```
demiankatz: Is there a roadmap for this functionality to become non-experimental? Right now this is triggering a "use at your own risk" warning.

demiankatz (commented):
This doesn't seem to break anything, but I'm a little concerned about using an experimental feature if we can avoid that (see below). I also think there are going to be some conflicts to resolve after you merge #118.

client/next.config.js
diboy2: I decided against this approach. I'm going to prevent incorrect import styles through eslint.

demiankatz (approved):
This makes sense to me -- thanks!

PR #123, opened 2022-04-04
Rename Fedora data classes for clarity.

demiankatz:
This PR renames the FedoraData model to FedoraDataCollection (to better differentiate it from the FedoraObject model, which represents a single object in Fedora as opposed to a collection of data pulled from Fedora) and renames the HierarchyCollector service to FedoraDataCollector (to correlate it with FedoraDataCollection, and to reflect the fact that it doesn't deal exclusively in hierarchies).

I believe this significantly improves the clarity of the naming in the backend code.

### DIFF
--- a/api/src/services/FedoraDataCollector.ts
+++ b/api/src/services/FedoraDataCollector.ts
@@ -31,7 +31,7 @@ class FedoraDataCollector {
         return FedoraDataCollector.instance;
     }
 
-    async getSingleObject(pid: string): Promise<FedoraDataCollection> {
+    async getObjectData(pid: string): Promise<FedoraDataCollection> {
         // Use Fedora to get data
         const DCPromise = this.fedora.getDublinCore(pid);
         const RDFPromise = this.fedora.getRdf(pid, false);
@@ -49,7 +49,7 @@ class FedoraDataCollector {
     }
 
     async getHierarchy(pid: string): Promise<FedoraDataCollection> {
-        const result = await this.getSingleObject(pid);
+        const result = await this.getObjectData(pid);
         // Create promises to retrieve parents asynchronously...
         const promises = (result.fedoraDetails.isMemberOf ?? []).map(async (resource) => {
             const parentPid = resource.split("/").pop();
--- a/api/src/services/FedoraDataCollector.ts
+++ b/api/src/services/FedoraDataCollector.ts
@@ -1,11 +1,11 @@
 import { Fedora } from "./Fedora";
 import Config from "../models/Config";
-import FedoraData from "../models/FedoraData";
+import FedoraDataCollection from "../models/FedoraDataCollection";
 import MetadataExtractor from "./MetadataExtractor";
 import TikaExtractor from "./TikaExtractor";
 
-class HierarchyCollector {
-    private static instance: HierarchyCollector;
+class FedoraDataCollector {
+    private static instance: FedoraDataCollector;
 
     fedora: Fedora;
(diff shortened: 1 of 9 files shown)


api/src/services/FedoraDataCollector.ts
```diff
@@ -48,8 +48,8 @@ class HierarchyCollector {
         );
     }
 
-    async getHierarchy(pid: string): Promise<FedoraData> {
-        const result = await this.getFedoraData(pid);
+    async getHierarchy(pid: string): Promise<FedoraDataCollection> {
+        const result = await this.getSingleObject(pid);
```
diboy2: What is this single object supposed to represent?

api/src/services/SolrIndexer.ts
```diff
@@ -110,8 +110,8 @@ class SolrIndexer {
 
         // Process parent data (note that hierarchyParents makes some special exceptions for VuFind;
         // fedoraParents exactly maintains the hierarchy as represented in Fedora):
-        const hierarchyParents: Array<FedoraData> = [];
-        const fedoraParents: Array<FedoraData> = [];
+        const hierarchyParents: Array<FedoraDataCollection> = [];
```
diboy2, line 113: What is the difference between these parents?

demiankatz: The comment above is intended to explain -- VuFind (our front end system) has a specific way of representing hierarchical data, so hierarchyParents follows VuFind's rules. The fedoraParents array is a raw representation of the data in Fedora, without applying VuFind-specific rules.

api/src/services/FedoraDataCollector.ts
demiankatz: Maybe `getSingleObject` is not the best name for the method -- I'm open to alternate suggestions. The idea here is that `getSingleObject` fetches all of the relevant data associated with a single object in Fedora, but does not retrieve data about its parents, while `getHierarchy` recursively calls `getSingleObject` to create a tree of parents and children.

demiankatz: Maybe `getDataCollectionForSingleObject` would be a more specific name.

api/src/services/SolrIndexer.ts
diboy2: What about vuFindParents?

api/src/services/FedoraDataCollector.ts
diboy2: getObjectDetails?

api/src/services/SolrIndexer.ts
demiankatz: I think renaming that variable could make sense, but I think it's beyond the scope of this PR since I don't want to change the internals of the classes at the same time that I'm renaming them. I will open a separate PR as a follow-up when this one is merged.

api/src/services/FedoraDataCollector.ts
demiankatz: How about getObjectData, since this is a DataCollector? It's a little shorter, and it makes the naming match up a little better.

diboy2 (approved):

api/src/services/FedoraDataCollector.ts
demiankatz: I've gone ahead and made this change. Thanks!

api/src/services/SolrIndexer.ts
demiankatz: See #125.

PR #124, opened 2022-04-06
Production changes. Generic Error. Object Editor Refactor

diboy2:
image


client/components/GenericError.tsx
```diff
@@ -0,0 +1,25 @@
...
+}
+
+const GenericError = ({ message }: GenericErrorProps) => {
+    return (
+        <Box display="flex" flexDirection="row" justifyContent="center" alignItems="center" minHeight="100vh">
+            <SentimentVeryDissatisfiedIcon sx={{ fontSize: "100px", color: "red" }} />
+            <br />
+            <Box width="300px">
+                <Box sx={{ fontSize: "24px", padding: "4px" }}>
+                    A generic error ocurred. Please return <Link href="/">home.</Link>
```
demiankatz: Typo: should be "occurred"

client/components/GenericError.test.tsx
```diff
@@ -0,0 +1,17 @@
...
+import GenericError from "./GenericError";
+
+describe("GenericError", () => {
+
+    it("renders", () => {
+        const wrapper = shallow(<GenericError message="Test error" />);
+        expect(toJson(wrapper)).toMatchSnapshot();
+    });
+
+    xit("shows an error message when pid is false", () => {
```
demiankatz: Unfinished test. (You probably are aware of this, but just want to make sure we have an open conversation as a reminder).

client/components/edit/ObjectEditor.tsx
```diff
@@ -1,5 +1,4 @@
 import React, { useEffect } from "react";
-import PropTypes from "prop-types";
```
demiankatz, line 2: Will we eventually be able to drop the prop-types dependency entirely, since we're leveraging Typescript instead now?

client/components/edit/__snapshots__/ChildList.test.tsx.snap
```diff
@@ -31,7 +31,7 @@ exports[`ChildList renders using ajax-loaded root data 1`] = `
   <ChildList>
     <ul>
       <li
-        key="rootchildfoo:124"
+        key="childfoo:124"
```
demiankatz: This key changed because the code currently expects pid to be null when we're accessing the root of the tree, but it is now blank string instead. I think some logic needs to be changed in ChildList.tsx around line 31 to make sure we don't have key conflicts.

client/util/routes.ts
```diff
@@ -10,34 +10,34 @@ const editObjectUrl = `${apiUrl}/edit/object`;
 const datastreamsUrl = `${editObjectUrl}/datastreams`;
 const editObjectCatalogUrl = `${apiUrl}/edit/catalog`;
 
-const getJobUrl = (category, job, extra = "") => {
+const getJobUrl = (category, job, extra = ""): string => {
```
demiankatz: All the parameters on many of these methods are strings; any reason they haven't been typed yet?

client/util/Zoomy.ts
```diff
@@ -1,3 +1,5 @@
+// @ts-nocheck
```
demiankatz, line 1: I agree that we don't want to go crazy converting this to fully typed Typescript... I think we may eventually want to replace this with an existing library like [OpenSeadragon](https://openseadragon.github.io/examples/tilesource-image/).

client/util/MagicLabeler.ts
```diff
@@ -1,3 +1,4 @@
+// @ts-nocheck
```
demiankatz: I think this code may stick around for a while, so it may be worth typing eventually. Maybe we should put a TODO comment on this as a reminder if it's too much work to do right now.

client/pages/edit/object/[pid]/newChild.tsx
```diff
@@ -1,9 +1,17 @@
...
+import CreateObject from "../../../../components/edit/create/CreateObject";
+import GenericError from "../../../../components/GenericError";
 
 export default function CreateObjectChild() {
     const router = useRouter();
     const { pid } = router.query;
-    return <CreateObject parentPid={pid} allowNoParentPid={false} allowChangeParentPid={false} />;
+    if (router.isReady) {
+        return !pid ? (
+            <GenericError message="allowChangeParentPid and allowNoParentPid cannot both be false when parentPid is empty." />
```
Comment (demiankatz) [inline client/pages/edit/object/[pid]/newChild.tsx]: Does it make sense to put this here? The original validation was inside the `<CreateObject>` component to prevent it from being used illegally. I understand the idea of displaying an error message in a better way, but in this scenario, will this error ever possibly trigger? Won't the route only match if pid is set to something, thus rendering the check unnecessary? Or am I overlooking an edge case?

client/next.config.js
```diff
@@ -1,4 +1,11 @@
 /** @type {import('next').NextConfig} */
 module.exports = {
   reactStrictMode: true,
+  typescript: {
+    // !! WARN !!
+    // Dangerously allow production builds to successfully complete even if
```
demiankatz, line 6: Is the goal to get enough typing in place here that we can remove this setting? If not, do we need to create an action to remove this in future?

client/context/EditorContext.tsx
```diff
@@ -1,7 +1,13 @@
+import { Snackbar } from "@mui/material";
 import React, { createContext, useContext, useReducer } from "react";
 import { editObjectCatalogUrl, getObjectModelsDatastreamsUrl } from "../util/routes";
 import { useFetchContext } from "./FetchContext";
 
+interface SnackbarState {
```
demiankatz, line 5: Is this the right place to define this? Is it used anywhere?

client/components/paginate/ZoomToggleButton.tsx
```diff
@@ -1,13 +1,12 @@
 import React from "react";
-import PropTypes from "prop-types";
 
-const ZoomToggleButton = ({ toggleZoom, zoom }) => {
-    return <button onClick={toggleZoom}>{`Turn Zoom ${zoom ? "Off" : "On"}`}</button>;
-};
+interface ZoomToggleButtonProps {
+    toggleZoom: void;
```
demiankatz: Should this type be `() => void` rather than just `void` since it's a void function, not a void value?

client/components/paginate/PaginatorControlGroup.tsx
```diff
@@ -1,7 +1,12 @@
 import React from "react";
-import PropTypes from "prop-types";
 
-const PaginatorControlGroup = ({ callback, children, label }) => {
+interface PaginatorControlGroupProps {
+    callback: void;
```
demiankatz: Another void function that maybe should be represented differently.

client/components/paginate/Thumbnail.tsx
```diff
@@ -1,8 +1,13 @@
 import React, { useEffect, forwardRef } from "react";
-import PropTypes from "prop-types";
 import { usePaginatorContext } from "../../context/PaginatorContext";
 
-const Thumbnail = forwardRef(({ scrollTo, selected, number }, ref) => {
+interface ThumbnailProps {
+    scrollTo: void;
```
demiankatz: Another void function. Apologies for raising this issue repeatedly if I'm wrong about it, but I think a change may be justified.

client/components/edit/create/CreateObject.tsx
```diff
@@ -0,0 +1,116 @@
...
+    const handleSubmit = async (event) => {
+        event.preventDefault();
+        const url = apiUrl + "/edit/object/new";
+        const data = {
+            title,
+            parent: parentPid,
+            model: selectedModel,
+            state: currentState,
+        };
+        console.log(data);
```
demiankatz: Is this console.log intentional?

client/components/edit/create/StateControls.tsx
```diff
@@ -0,0 +1,31 @@
+import React, { Dispatch } from "react";
+
+interface StateControlsProp {
```
demiankatz: We're eventually going to have an "edit state" dialog box, so I wonder if StateControls should be in a more generic location than /create.

client/components/edit/create/ParentControls.tsx
```diff
@@ -0,0 +1,82 @@
+import React, { useState } from "react";
+
+interface ParentControlProps {
+    allowChangeParentPid: boolean;
+    parent: string;
+    handleParentChange: React.ChangeEventHandler<HTMLInputElement> | undefined;
+    noParent: boolean;
+}
+
+const ParentControl = ({ allowChangeParentPid, parent, handleParentChange, noParent }: ParentControlProps) => {
```
demiankatz: Like StateControl, this may need to be reused in future in an editing rather than creation mode, so maybe it should go somewhere else. It will also eventually need to be improved to incorporate the ability to pick the parent PID out of a tree as an alternative to hand-entering it (though that's not top priority right now).

client/components/edit/create/__snapshots__/CreateObject.test.tsx.snap
```diff
@@ -19,7 +19,7 @@ exports[`CreateObject renders appropriately with default settings 1`] = `
       name="parent"
       required=""
       type="text"
-      value=""
+      value="foo:1234"
```
demiankatz: foo:1234 is not a default setting -- is this test passing more parameters than intended?

client/hooks/useJob.ts
```diff
@@ -53,7 +53,14 @@ const useJob = ({ category, children }) => {
             ...getPublishedStatusText({ derivatives, ingesting, published }),
         ];
     };
-    const getStatusText = ({ derivatives, minutes_since_upload, documents, audio, ingesting, published }) => {
+    const getStatusText = ({
```
demiankatz, line 67: Does this still need to be typed? It just seems to have been reformatted so far.

client/hooks/useJob.ts
```diff
@@ -109,10 +116,7 @@ const useJob = ({ category, children }) => {
         await updateStatus();
     };
 
-    const updateStatus = async (e) => {
-        if (typeof e !== "undefined") {
-            e.stopPropagation();
-        }
+    const updateStatus = async () => {
```
demiankatz: Any idea why this event code was here to begin with? Just want to be sure I understand why it's being removed.

demiankatz (commented):
Thanks, @diboy2, this looks like a very solid start. As usual, I have a bunch of comments, but most are very minor.

I'd also suggest in future that something like this might work better as multiple pull requests -- there are a lot of unrelated things going on in here -- but I understand how these changes could have easily gotten entangled, and at this early phase of the project, I'm less concerned about the elegance of our commit history. :-)

client/components/edit/ObjectEditor.tsx
diboy2: Yes

demiankatz: I'll leave this comment thread open as a reminder to remove it from package.json at the end of the process.

demiankatz: ...or, then again, maybe it's an indirect dependency and there's no action needed. Should have checked first! :-)

client/components/GenericError.tsx
diboy2: Made change

client/components/edit/__snapshots__/ChildList.test.tsx.snap
diboy2: Made change.

client/util/routes.ts
diboy2: Made change.

client/util/Zoomy.ts
diboy2: Okay.

client/components/edit/create/CreateObject.tsx
diboy2: Made change.

client/components/edit/create/ParentControls.tsx
diboy2: Made change.

client/components/edit/create/StateControls.tsx
diboy2: Made change

client/pages/edit/object/[pid]/newChild.tsx
Comment (diboy2) [reply client/pages/edit/object/[pid]/newChild.tsx]: I want ahead and tested this scenario and it looks like this scenario would never happen, so I'm going to go ahead and remove this and delete the associated the unit test.

client/next.config.js
diboy2: Yes. The goal is to put the majority of typing in place. So, we should create an action to remove this one day.

client/components/paginate/ZoomToggleButton.tsx
diboy2: Made change.

client/components/paginate/PaginatorControlGroup.tsx
diboy2: Made change.

client/components/paginate/Thumbnail.tsx
diboy2: Made change.

client/hooks/useJob.ts
diboy2: I'm going to reviert the change.

client/util/MagicLabeler.ts
diboy2: Added a todo.

client/next.config.js
demiankatz: I've added a card to our Teams board.

client/components/GenericError.test.tsx
diboy2: I think the first test should be sufficient. Im going to add a check that the message exists.

client/components/edit/create/__snapshots__/CreateObject.test.tsx.snap
diboy2: Reverted change.

client/hooks/useJob.ts
diboy2: Made change.

client/context/EditorContext.tsx
demiankatz: I think this thread may have gotten hidden by GitHub's collapsed conversations feature -- I believe the question remains open.

client/components/edit/create/__snapshots__/CreateObject.test.tsx.snap
demiankatz: In addition to reverting the change to the snapshot, I think it probably makes sense to pass `{}` instead of `props` to `CreateObject` in line 68 of CreateObject.test.ts (that may be a pre-existing bug... but the intent was to test how it behaves if you just rely on the default values provided by the function).

client/hooks/useJob.ts
demiankatz: Thanks! Just one syntax question: I see that you defined an interface to solve this, and all the fields of the interface are comma-separated. I thought Typescript interfaces usually had a semi-colon at the end of every line. The linter isn't complaining, but I'm curious if this needs to be changed.

diboy2: Weird. I was using the api for reference. I'll make the change.

client/components/edit/create/__snapshots__/CreateObject.test.tsx.snap
diboy2: Made change.

client/components/edit/create/__snapshots__/CreateObject.test.tsx.snap
```diff
@@ -67,11 +67,15 @@ exports[`CreateObject renders appropriately with editing disabled 1`] = `
...
     />
   </label>
-  <input
-    name="parent"
-    type="hidden"
-    value="foo:1234"
-  />
+  <label>
+    Parent ID
+    <input
```
demiankatz: If editing is disabled, I think this should render as a hidden element rather than a text input. Did something change?

client/components/edit/ChildList.tsx
```diff
@@ -1,19 +1,22 @@
...
+interface ChildListProps {
+    pid: string;
+}
+
+const ChildList = ({ pid = "" }: ChildListProps) => {
     const {
         action: { fetchJSON },
     } = useFetchContext();
-    const [children, setChildren] = useState([]);
+    const [children, setChildren] = useState<any[]>([]);
```
demiankatz: Can/should we be more specific than `any` here? What is the return type of fetchJSON? Can/should we use that?

client/components/paginate/PaginatorControlGroup.tsx
```diff
@@ -1,7 +1,12 @@
 import React from "react";
-import PropTypes from "prop-types";
 
-const PaginatorControlGroup = ({ callback, children, label }) => {
+interface PaginatorControlGroupProps {
+    callback: React.MouseEventHandler<HTMLButtonElement> | undefined;
```
demiankatz: Why is undefined allowed?

client/components/paginate/ZoomToggleButton.tsx
```diff
@@ -1,13 +1,12 @@
 import React from "react";
-import PropTypes from "prop-types";
 
-const ZoomToggleButton = ({ toggleZoom, zoom }) => {
-    return <button onClick={toggleZoom}>{`Turn Zoom ${zoom ? "Off" : "On"}`}</button>;
-};
+interface ZoomToggleButtonProps {
+    toggleZoom: React.MouseEventHandler<HTMLButtonElement> | undefined;
```
demiankatz: Why is undefined allowed?

client/pages/edit/object/[pid]/newChild.tsx
Comment (demiankatz) [reply client/pages/edit/object/[pid]/newChild.tsx]: Looks like this didn't get removed.

demiankatz (changes requested):
Sorry, took another look and found a few more small things...

client/components/paginate/PaginatorControlGroup.tsx
diboy2: In javascript, a function without a return type returns undefined. I'll change it to void since typescript offers that type.

demiankatz: Makes sense; thanks for the clarification!

client/components/paginate/ZoomToggleButton.tsx
diboy2: Made change.

client/components/edit/create/__snapshots__/CreateObject.test.tsx.snap
diboy2: I didn't see the need to use a hidden element, since that value is already stored in a react state.

demiankatz: Good point regarding the lack of need for a hidden input... but in any case, the intention is that the visible parent element should not display when parent editing is disabled. If that became visible unintentionally as a result of refactoring, further adjustment may be needed. (I haven't done extensive hands-on testing around this particular issue, though, so maybe I'm misreading the changes in the snapshots... I can spend time on that if you need me to).

client/components/edit/ChildList.tsx
diboy2: Made change.

client/components/edit/create/__snapshots__/CreateObject.test.tsx.snap
diboy2: I made the adjustments. That was the bug I was talking about.

client/components/edit/datastream/DatastreamControlButton.tsx
```diff
@@ -16,7 +15,17 @@ const Icons = {
     Delete: <Delete />,
 };
 
-const DatastreamControlButton = ({ modalState, datastream, disabled }) => {
+interface DatastreamControlButtonProps {
+    modalState: string;
+    disabled: boolean;
```
demiankatz: I realize that order doesn't really matter, but I wonder if it would be more clear/readable if the order of the interface and the order of the extracted parameters matched up. For some reason, disabled and datastream are transposed here.

client/components/paginate/Thumbnail.tsx
```diff
@@ -1,8 +1,13 @@
 import React, { useEffect, forwardRef } from "react";
-import PropTypes from "prop-types";
 import { usePaginatorContext } from "../../context/PaginatorContext";
 
-const Thumbnail = forwardRef(({ scrollTo, selected, number }, ref) => {
+interface ThumbnailProps {
+    scrollTo: () => void;
+    number: number;
```
demiankatz, line 6: This is another instance where the orders are transposed (in this case, selected vs. number).

client/components/edit/create/CreateObject.tsx
```diff
@@ -0,0 +1,138 @@
...
+                    parent,
+                    model: selectedModel,
+                    state: currentState,
+                }),
+            };
+            const headers = { "Content-Type": "application/json" };
+            const pid = await fetchText(newEditObjectUrl, params, headers);
+            setResults({
+                state: "loading",
+                content: pid,
```
demiankatz, line 82: The original CreateObject component displayed an "object created" message followed by a link to the newly-created PID upon success. The refactored component just outputs the PID by itself in an unlinked state.

I would definitely like to see the UI improved so that CreateObject opens in a lightbox and reports success through the SnackBar. Is this an incremental step in that direction, or an oversight? I'm nearly ready to approve this PR, but this feels like a loss of functionality. If there's not a simple way to fix it here, I'm happy to merge as-is if we immediately follow up with another PR to restore the functionality in an improved way. What do you think?

demiankatz (changes requested):
Nearly done, I think -- just two nitpicks and one more significant issue.

client/components/edit/datastream/DatastreamControlButton.tsx
diboy2: Made change

client/components/paginate/Thumbnail.tsx
diboy2: Made change.

client/components/edit/create/CreateObject.tsx
diboy2: I fixed the issue with object editor success state. Atm, the snackbar component was isolated to the datastream features. If we want to utilize the Snackbar feature for the whole application, I would have to make some signficant modifications. Could we create a story for that task?

demiankatz: Thanks for the fix. Do we need test coverage for this case?

Good idea regarding adding a story related to Snackbar refactoring. I've added one in Teams. I wouldn't give that high urgency yet, though -- but a reusable Snackbar could definitely be considered a future "quality of life" thing.

demiankatz (approved):
I did have an opportunity to test this out, and it seems to be working as expected now. Feel free to merge, though if you have a moment to add another test to cover the situation of successful object creation generating an appropriate link, that might be worth adding to protect against regressions in that area.

client/components/edit/create/CreateObject.tsx
diboy2: Added unit tests.

PR #127, opened 2022-04-12
Include dc:identifier in initial object metadata.

demiankatz:
While testing, I noticed an oversight: the code was not initializing DC metadata with the appropriate identifier value. This PR corrects the problem.


diboy2 (approved):

demiankatz:
Thanks, @diboy2 -- merging now.

PR #128, opened 2022-04-19
Improved routes

demiankatz:
This pull request standardizes the routes in the API for consistency:

- Object-related routes now take the form `/object/[pid]/action`
- The datastream upload route now uses `/object/[pid]/datastream/[stream]` for consistency with other stream-related routes
- The client route methods now use helper methods to enforce consistent application of patterns
- All pid and stream values are now properly URI encoded for safety

### DIFF
--- a/api/src/routes/edit.ts
+++ b/api/src/routes/edit.ts
@@ -120,8 +120,9 @@ function uploadFile(req, res, next) {
         }
     });
 }
+edit.post("/object/:pid", requireToken, pidSanitizer, uploadFile);
 
-edit.get("/object/modelsdatastreams/:pid", requireToken, pidSanitizer, async function (req, res) {
+edit.get("/object/:pid/modelsdatastreams", requireToken, pidSanitizer, async function (req, res) {
     try {
         const data = await FedoraDataCollector.getInstance().getObjectData(req.params.pid);
         res.json({ models: data.models, datastreams: data.fedoraDatastreams });
@@ -130,10 +131,9 @@ edit.get("/object/modelsdatastreams/:pid", requireToken, pidSanitizer, async fun
         res.status(500).send(error.message);
     }
 });
-edit.post("/object/:pid", requireToken, pidSanitizer, uploadFile);
-edit.get("/object/children", requireToken, getChildren);
-edit.get("/object/children/:pid", requireToken, pidSanitizer, getChildren);
-edit.get("/object/details/:pid", requireToken, pidSanitizer, async function (req, res) {
+edit.get("/treeTop", requireToken, getChildren);
+edit.get("/object/:pid/children", requireToken, pidSanitizer, getChildren);
+edit.get("/object/:pid/details", requireToken, pidSanitizer, async function (req, res) {
     const pid = req.params.pid;
     const obj = FedoraObject.build(pid);
     const sort = await obj.getSort();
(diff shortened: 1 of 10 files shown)


demiankatz:
One outstanding question that I did not attempt to address here: we currently have `/models` and `/catalog/models` routes which do slightly different things. As far as I can tell, `/catalog/models` is not used. Should we move the current `/models` functionality to `/catalog/models` or is there a reason to keep these separate? It seems to me it would be better to unify catalog-related API methods.

api/src/routes/edit.ts
```diff
@@ -120,8 +119,9 @@ function uploadFile(req, res, next) {
         }
     });
 }
+edit.post("/object/:pid/datastream/:stream", requireToken, datastreamSanitizer, uploadFile);
```
diboy2, line 122: Oh yeah. Thanks for catching this.

api/src/routes/edit.ts
```diff
@@ -130,10 +130,9 @@ edit.get("/object/modelsdatastreams/:pid", requireToken, pidSanitizer, async fun
         res.status(500).send(error.message);
     }
 });
-edit.post("/object/:pid", requireToken, pidSanitizer, uploadFile);
-edit.get("/object/children", requireToken, getChildren);
-edit.get("/object/children/:pid", requireToken, pidSanitizer, getChildren);
-edit.get("/object/details/:pid", requireToken, pidSanitizer, async function (req, res) {
+edit.get("/treeTop", requireToken, getChildren);
```
diboy2: is this supposed to be hierarchy tree? or root of the tree?

client/util/routes.ts
```diff
@@ -30,24 +30,44 @@ const getStatusUrl = (category: string, children: string): string => {
...
-    return `${editObjectUrl}/modelsdatastreams/${pid}`;
+    return getPidActionUrl(pid, "modelsdatastreams");
+}
+
+const getObjectParentsUrl = (pid: string): string => {
+    return getPidActionUrl(pid, "parents");
+}
+
+const getDatastreamActionUrl = (pid: string, datastream: string, action = ""): string => {
+    return getPidActionUrl(pid, `datastream/${encodeURIComponent(datastream)}` + (action.length > 0 ? `/${action}` : ""));
```
diboy2, line 54: if the string is empty, wouldn't this still call the same url regardless if the backslash exists or not?

api/src/routes/edit.ts
demiankatz: It's just listing the children of the root of the tree -- top-level PIDs. I'm not attached to calling it treeTop. Could be treeRoot, or any better name you care to suggest.

client/util/routes.ts
demiankatz: Yes, that's true. We could probably simplify this a little bit, though it just feels "cleaner" to me to omit the slash when we don't need it.

api/src/routes/edit.ts
diboy2: Maybe topLevelTree?

demiankatz: Thanks for the suggestion. Since this doesn't really directly relate to the entire tree, I opted to change it to topLevelObjects -- I think that's the most descriptive name. I've just pushed up a commit. What do you think?

diboy2: I think that's a closer description.

client/util/routes.ts
diboy2: ok

diboy2 (approved):

demiankatz:
Thanks, @diboy2 -- merging now!

PR #132, opened 2022-04-21
Add pagination control to child lists.

demiankatz:
I'm planning on enhancing child lists in the near future -- right now, they are only displaying raw data from the Solr index, but I plan to add more detailed child components that pull additional data from other places (e.g. thumbnails from Fedora). Since loading extra data will be more expensive, we can't have hundreds or thousands of them displaying at once -- thus, I am introducing a pagination component to constrain the scope.


diboy2 (approved):

demiankatz:
Thanks, @diboy2!

PR #133, opened 2022-04-21
Fix display order of children.

demiankatz:
The child routes displayed children in arbitrary order; this PR fixes them so that they use parent-specific sort order (when applicable) with title as a tie-breaker.


diboy2 (approved):

demiankatz:
Thanks, @diboy2 -- merging now.

PR #134, opened 2022-04-22
Refactor database functionality; add MySQL support

demiankatz:
This PR refactors database code to make it easier to test and manage, and adds MySQL compatibility and the ability to control database settings via vudl.ini. It also changes date processing to make it more platform-agnostic.

TODO
- [x] Add test coverage


api/src/services/Database.ts
```diff
@@ -24,126 +20,168 @@ interface Pid {
...
 
-        if (REBUILD_DB || !fs.existsSync(dbFilename)) {
+        await this.initialize(this.connection);
+
+        console.log("Database:ready");
+    }
+
+    protected async initialize(db): Promise<void> {
+        // Special case: if we're using a disk-based database, make sure the containing directory exists:
+        if (this.config.databaseClient === "sqlite3") {
```
diboy2, line 58: Is there a reason to keep sqlite3 if we're using mysql?

demiankatz: Since we're going to be open-sourcing this tool, I think it's useful to offer multiple options... and sqlite is a good tool for evaluating the software, since it requires less setup to get up and running quickly. That being said, I don't think it's critically important, but I figure it doesn't hurt to keep it working if it's not too expensive to do so.

demiankatz:
I've just added a test suite for the new Database class. This only gets us to about 50% coverage, but it's significantly better than nothing, and it covers the most potentially dangerous behavior. I'll try to flesh it out further as time permits, but I don't think we need to hold off merging until that is done -- I'd prefer to make these improvements available as soon as possible. Any objections or further feedback, @diboy2?

diboy2 (approved):

PR #135, opened 2022-04-25
Clear config for tests.

demiankatz:
I noticed that some tests were complaining about a missing vudl.ini in CI, but we don't want tests to try to access real configuration files. This explicitly sets the configuration to empty for the offending tests.


diboy2 (approved):

demiankatz:
Thanks, @diboy2!

PR #136, opened 2022-04-26
Authentication improvements

demiankatz:
This PR refactors and improves authentication logic, adding support for more methods.

TODO
- [x] Eliminate hard-coded user accounts in code
- [x] Make database-driven username/password login work properly
- [x] Polish database login (add salt, etc.)
- [x] Add SAML support
- [x] Add "no login required" option
- [x] Add "allowed users" configuration list
- [x] Add test coverage


demiankatz:
@diboy2 / @crhallberg, much as I'm going to miss the "just log in as chris" button, this PR sets up all the real authentication options we are going to need for the system: SAML integration for production use, plus the ability to use a local username/password database (or just usernames without passwords) for testing purposes.

I still need to improve test coverage, but I'd love to get some feedback on this before I set it all in stone with tests.

api/src/routes/auth.ts
```diff
@@ -93,28 +64,40 @@ router.use(function (req, res, next) {
...
-    res.render("../../views/login-test", { user });
-});
+    next();
+}
 
-router.get("/logout", function (req, res) {
-    req.logout();
-    res.redirect(Config.getInstance().clientUrl);
-});
+function showLoginForm(req, res) {
```
diboy2, line 68: Does this login form need to live in the api folder?

demiankatz: Because it is used to establish a session on the API server, which is independent of any client that might be consuming the API, I think it makes sense to keep it on the API side.

demiankatz: (Essentially this is the service that the React client delegates to in order to perform a login, when it hasn't yet acquired a token).

diboy2 (approved):

demiankatz:
Thanks for the review, @diboy2 -- I've added a bit more test coverage. There's still much to do, but I believe we're leaving with more coverage (and more testable code) than we came in with, so it's progress.

PR #137, opened 2022-04-29
License control feature

diboy2:


api/src/models/FedoraObject.ts
```diff
@@ -130,6 +130,18 @@ export class FedoraObject {
         );
     }
 
+    async modifyLicense(stream: string, licenseKey: string) {
+        const licenses = this.config.licenses;
+        const url = licenses[licenseKey];
+        const licenseXml = `
+            <METS:rightsMD xmlns:METS="http://www.loc.gov/METS/" ID="0">
+                <METS:mdRef xmlns:xlink="http://www.w3.org/1999/xlink" LOCTYPE="URL" MDTYPE="OTHER" MIMETYPE="text/html" OTHERMDTYPE="HTML" xlink:href="${url}">
```
demiankatz: It would probably be a good idea to call `xmlescape` on the URL in case it contains special characters like `&`. See [here](https://github.com/FalveyLibraryTechnology/VuDLPrepJS/blob/d3ef26cdf66a3d87dc249a80364ef0c7787da810/api/src/services/Fedora.ts#L380) for an example.

api/src/routes/edit.test.ts
```diff
@@ -99,6 +99,65 @@ describe("edit", () => {
         });
     });
 
+    describe("post /object/:pid/datastream/:stream/license", () => {
+        let datastreamManager;
+        let licenseKey;
+        beforeEach(() => {
```
demiankatz, line 103: Might be a good idea to have an `afterEach` to reset the mocks.

api/src/routes/edit.ts
```diff
@@ -126,6 +126,58 @@ function uploadFile(req, res, next) {
...
+            await datastream.uploadLicense(pid, stream, licenseKey);
+            res.status(200).send("Upload license success");
+        } catch (error) {
+            console.log("error", error);
+            res.status(500).send(error.message);
+        }
+    }
+);
+
+edit.get("/object/:pid/datastream/:stream/license", requireToken, datastreamSanitizer, async (req, res) => {
```
demiankatz, line 148: You seem to define this same route twice for some reason.

api/src/routes/edit.ts
```diff
@@ -126,6 +126,58 @@ function uploadFile(req, res, next) {
...
+edit.post(
+    "/object/:pid/datastream/:stream/license",
+    requireToken,
+    bodyParser.json(),
+    datastreamSanitizer,
+    async function (req, res) {
+        const { pid, stream } = req.params;
+        const { licenseKey } = req.body;
+        try {
+            console.log("Uploading license", licenseKey);
```
demiankatz: Do we need this console.log in the long term, or is this just for testing?

client/components/edit/datastream/DatastreamLicenseContent.tsx
```diff
@@ -0,0 +1,74 @@
...
+import FormControl from "@mui/material/FormControl";
+import FormControlLabel from "@mui/material/FormControlLabel";
+import FormLabel from "@mui/material/FormLabel";
+import Radio from "@mui/material/Radio";
+import RadioGroup from "@mui/material/RadioGroup";
+import useDatastreamOperation from "../../../hooks/useDatastreamOperation";
+import { useEditorContext } from "../../../context/EditorContext";
+
+const licenseMapping = {
+    ATTRIBUTION: "Attribution",
```
demiankatz: The intention is to be able to add licenses purely via configuration, so if we have constants in the config file and then map them to strings in the code, this is less flexible than intended. Can we just have a map of description => url (or vice versa) in the .ini file, or are the constraints of the .ini format too strict to allow that to work?

demiankatz: Maybe the solution is as simple as moving this map into the config file as well -- I do see advantages to using tokens for licenses (particularly if we need to add an i18n layer later), but we should find a solution where this isn't hard-coded into the components.

client/components/edit/datastream/DatastreamUploadModalContent.tsx
```diff
@@ -1,18 +1,27 @@
...
-        <DialogContent>
-            <input className="uploadFileButton" type="file" onChange={onChange} />
-        </DialogContent>
-    );
+    switch (activeDatastream) {
+        case "LICENSE":
+            return <DatastreamLicenseContent />;
+        default:
+            return (
+                <DialogContent>
```
demiankatz: Would it make sense to refactor this to `<DatastreamUploadContent />`? I realize it's short, but that might make sense in the long run (but no pressure if you prefer to leave it as is for now).

demiankatz (changes requested):
@diboy2, I haven't had a chance to do any hands-on testing yet, but I did a quick initial review -- see a few comments below. Thanks for the progress; as always, this looks to be off to an excellent start!

demiankatz:
@diboy2, I also noticed that my recent work introduced a minor conflict; I've gone ahead and resolved it for you.

api/src/services/DatastreamManager.test.ts
```diff
@@ -181,6 +191,60 @@ describe("DatastreamManager", () => {
...
+        beforeEach(() => {
+            pid = "test1";
+            stream = "test2";
+            licenseKey = "test3";
+        });
+
+        it("calls uploadLicense with success", async () => {
+            modifyLicenseSpy.mockResolvedValue("");
+
+            await datastreamManager.uploadLicense(pid, stream, licenseKey);
```
demiankatz, line 217: Should we have an expectation here to indicate that the pid is getting used correctly? Is it just a matter of spying on FedoraObject.build? Same question goes for a few other tests below that include pid values but don't verify them.

api/src/models/FedoraObject.ts
diboy2: Made change.

api/src/services/DatastreamManager.test.ts
```diff
@@ -181,6 +191,60 @@ describe("DatastreamManager", () => {
...
+            extractLicenseSpy.mockReturnValue("testLicense");
+
+            const response = await datastreamManager.getLicenseKey(pid, stream);
+
+            expect(getDatastreamSpy).toHaveBeenCalledWith(stream);
+            expect(extractLicenseSpy).toHaveBeenCalledWith("testXml");
+            expect(response).toEqual("testLicenseKey");
+        });
+
+        it("returns an empty string", async () => {
```
demiankatz: Maybe expand the name of this test to "returns an empty string when an unrecognized URI is provided".

api/src/routes/edit.test.ts
diboy2: Made change.

api/src/routes/edit.ts
diboy2: Made change.

client/components/edit/datastream/__snapshots__/DatastreamUploadModalContent.test.tsx.snap
```diff
@@ -9,3 +9,5 @@ exports[`DatastreamUploadModalContent renders 1`] = `
   />
 </ForwardRef(DialogContent)>
 `;
+
+exports[`DatastreamUploadModalContent renders DatastreamLicenseContent 1`] = `null`;
```
demiankatz: Should this snapshot really be null?

client/components/edit/datastream/DatastreamLicenseContent.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+            uploadLicense: jest.fn(),
+            getLicenseKey: jest.fn(),
+        };
+        mockUseEditorContext.mockReturnValue(editorValues);
+        mockUseDatastreamOperation.mockReturnValue(datastreamOperationValues);
+    });
+
+    it("renders", () => {
+        const wrapper = mount(<DatastreamLicenseContent />);
+        expect(toJson(wrapper)).toMatchSnapshot();
```
demiankatz, line 46: Would it make sense to mock some of the MUI components in this test to make the snapshot more concise/readable? As it is, the snapshot is incredibly huge. I realize that snapshots aren't necessarily intended to be human-readable, but it still might be worth making it a little more review-friendly.

client/hooks/useDatastreamOperation.test.ts
```diff
@@ -108,6 +111,55 @@ describe("useDatastreamOperation", () => {
         });
     });
 
+    describe("uploadLicense", () => {
+        beforeEach(() => {
+            editorValues.state.activeDatastream = "LICENSE";
+        });
+
+        it("uploades a license with success", async () => {
```
demiankatz: Typo: "uploades" should be "uploads"

demiankatz (changes requested):
Thanks for the latest progress -- here are a few comments on the unit tests.

api/src/models/Config.ts
```diff
@@ -219,6 +219,8 @@ class Config {
 
     get samlEntryPoint(): string {
         return (this.authenticationSettings["saml_entry_point"] as string) ?? "";
+    get licenses(): Record<string, string> {
```
demiankatz: There appears to be a merge error here -- two functions have gotten incorrectly combined. You should just need to insert a close brace and newline above `get licenses()` to fix it.

client/components/edit/datastream/DatastreamLicenseContent.tsx
diboy2: I think I'll add it to the catalog call.

api/src/routes/edit.ts
diboy2: One of them is a "GET"  call that reads the particular license. The other one is a "POST" call that updates the particular license.

demiankatz: Are you sure? I see two identical `edit.get("/object/:pid/datastream/:stream/license"` definitions here...

demiankatz: (Well, the code isn't identical, but the declarations are... If one is supposed to be a POST, shouldn't it be `edit.post`?)

api/src/services/DatastreamManager.test.ts
diboy2: Made change.

client/hooks/useDatastreamOperation.test.ts
diboy2: Made change.

client/components/edit/datastream/DatastreamLicenseContent.tsx
demiankatz: Sounds reasonable to me.

api/src/routes/edit.ts
diboy2: Oh. hmm. Let me check.

diboy2: Okay. I removed it.

api/src/routes/edit.ts
```diff
@@ -126,6 +126,46 @@ function uploadFile(req, res, next) {
...
+        const { pid, stream } = req.params;
+        const datastream = DatastreamManager.getInstance();
+        const licenseKey = await datastream.getLicenseKey(pid, stream);
+        res.status(200).send(licenseKey);
+    } catch (error) {
+        res.status(500).send(error.message);
+    }
+});
+
+edit.get("/object/:pid/datastream/:stream/metadata", requireToken, datastreamSanitizer, async (req, res) => {
```
demiankatz, line 159: Is this metadata route used for anything yet? I see we already defined `getObjectDatastreamMetadataUrl` in a previous pull request but haven't used it yet... and this seems related to that, but also not used yet. Should we move this to a separate PR for now, or is it part of the current work and I'm just missing something?

demiankatz (commented):
Thanks for the progress on the review comments -- just one more small question from my latest review.

api/src/services/DatastreamManager.test.ts
diboy2: made change

client/components/edit/datastream/__snapshots__/DatastreamUploadModalContent.test.tsx.snap
diboy2: Made change

api/src/routes/edit.ts
diboy2: I just added since it was useful for testing. Even though it's not in it's final form, I figured I might as well add it since getting the metadata is going to be part of future work anyway. Let me know if it produces a potential problem.

demiankatz: As long as it's intentional, I don't object -- just wanted to be sure it wasn't an artifact that got added by accident. :-)

demiankatz (approved):
This is looking great to me now. Thanks again!

PR #138, opened 2022-05-02
Improve object summary

demiankatz:
This PR is an attempt to make object data retrieval more consolidated and less redundant. It combines two distinct API calls into one since they make overlapping Fedora calls, and it revises the ObjectSummary component to leverage the EditorContext.

### DIFF
--- a/api/src/models/FedoraDataCollection.ts
+++ b/api/src/models/FedoraDataCollection.ts
@@ -172,7 +172,7 @@ class FedoraDataCollection {
         return this.fedoraDetails.sequence ?? [];
     }
 
-    get sort(): string {
+    get sortOn(): string {
         if ((this.fedoraDetails?.sortOn ?? []).length > 0) {
             return this.fedoraDetails.sortOn[0];
         }
--- a/api/src/models/FedoraDataCollection.ts
+++ b/api/src/models/FedoraDataCollection.ts
@@ -172,6 +172,13 @@ class FedoraDataCollection {
         return this.fedoraDetails.sequence ?? [];
     }
 
+    get sort(): string {
+        if ((this.fedoraDetails?.sortOn ?? []).length > 0) {
+            return this.fedoraDetails.sortOn[0];
+        }
+        return "title";
+    }
+
     get title(): string {
         return (this.metadata["dc:title"] ?? [])[0] ?? "";
     }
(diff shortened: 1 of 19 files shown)


api/src/routes/edit.ts
```diff
@@ -127,25 +124,24 @@ function uploadFile(req, res, next) {
...
 
-edit.get("/object/:pid/modelsdatastreams", requireToken, pidSanitizer, async function (req, res) {
+edit.get("/topLevelObjects", requireToken, getChildren);
+edit.get("/object/:pid/children", requireToken, pidSanitizer, getChildren);
+edit.get("/object/:pid/details", requireToken, pidSanitizer, async function (req, res) {
     try {
-        const data = await FedoraDataCollector.getInstance().getObjectData(req.params.pid);
-        res.json({ models: data.models, datastreams: data.fedoraDatastreams });
+        const pid = req.params.pid;
+        const data = await FedoraDataCollector.getInstance().getObjectData(pid);
```
diboy2: Since you already surround this with a try catch, you can just do `const { fedoraDatastreams, metadata, models, sort }`

api/src/models/FedoraDataCollection.ts
```diff
@@ -172,6 +172,13 @@ class FedoraDataCollection {
         return this.fedoraDetails.sequence ?? [];
     }
 
+    get sort(): string {
```
diboy2: Should be this renamed to sortOn?

api/src/routes/edit.ts
demiankatz: Thanks! Implemented in [af0732f](https://github.com/FalveyLibraryTechnology/VuDLPrepJS/pull/138/commits/af0732fddc17d9489fe7ae37f492818a23a782c9).

api/src/models/FedoraDataCollection.ts
demiankatz: I'm not sure if it makes a huge difference, but for the sake of consistency and specificity, I went ahead and made the change: [5369e9d](https://github.com/FalveyLibraryTechnology/VuDLPrepJS/pull/138/commits/5369e9dfdbe97e00ebe42c5887f1f6f68ee34bf0). Thanks!

diboy2 (approved):

PR #140, opened 2022-05-05
Refactor ChildList component and add functionality

demiankatz:
This PR refactors the ChildList component and adds functionality to Child displays.

TODO
- [x] Create separate directory for child-related components
- [x] Separate Child component from ChildList
- [x] Load details directly from Fedora instead of relying on Solr, to ensure up-to-date displays.
- [x] Add ability to expand sub-trees of children
- [x] Eliminate redundant extractMetadata function in Child component
- [x] Add test coverage for Child component
- [x] Add test coverage for ChildListContext

Future tasks (for subsequent PR(s)):
- Display activation status on children
- Display sort order on children (and allow it to be edited)
- Display additional details/statistics as appropriate

### DIFF
--- a/client/components/edit/children/ChildList.tsx
+++ b/client/components/edit/children/ChildList.tsx
@@ -28,9 +28,9 @@ export const ChildList = ({ pid = "", pageSize = 10 }: ChildListProps): React.Re
     const childDocs = children.docs;
     const contents =
         childDocs.length > 0 ? (
-            childDocs.map((child) => {
+            childDocs.map((child: Record<string, string>) => {
                 return (
-                    <li key={(pid || "root") + "child" + child.id}>
+                    <li key={`${pid}_child_${child.id}`}>
                         <Child pid={child.id} initialTitle={child.title ?? "-"} />
                     </li>
                 );
--- a/client/components/edit/children/ChildList.tsx
+++ b/client/components/edit/children/ChildList.tsx
@@ -38,13 +38,15 @@ export const ChildList = ({ pid = "", pageSize = 10 }: ChildListProps): React.Re
         }
         loadData();
     }, [page]);
-    const contents = (children?.docs ?? []).map((child) => {
-        return (
-            <li key={(pid || "root") + "child" + child.id}>
-                <Child pid={child.id} initialTitle={child.title ?? "-"} />
-            </li>
-        );
-    });
+    const childDocs = children?.docs ?? [];
+    const contents = childDocs.length > 0
+        ? (childDocs).map((child) => {
+            return (
+                <li key={(pid || "root") + "child" + child.id}>
+                    <Child pid={child.id} initialTitle={child.title ?? "-"} />
+                </li>
+            );
+        }) : <p>Empty.</p>;
     const pageCount = Math.ceil(children.numFound / pageSize);
     const paginator =
         pageCount > 1 ? (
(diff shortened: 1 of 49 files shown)


client/context/ChildListContext.tsx
```diff
@@ -0,0 +1,103 @@
...
+    start?: number;
+    docs?: Record<string, string>[];
+}
+
+interface ChildListState {
+    childStorage: Record<string, Children>;
+}
+
+const childListContextParams: ChildListState = {
+    childStorage: {},
```
diboy2: Could we use a typescript Map here? https://howtodoinjava.com/typescript/maps/

demiankatz: That does seem like a very good fit. My only question is how should we handle this in the reducer? Do we need to be concerned about mutability? That is, can we just mutate the Map in the state and then return the state, or is it strictly necessary to create a fresh copy of the whole state? (I've been assuming the latter to be the case, but maybe I'm wrong).

client/components/edit/children/ChildList.tsx
```diff
@@ -1,52 +1,43 @@
...
+    if (notLoaded) {
+        return <p>Loading...</p>;
+    }
+    const children = childStorage[key];
+    const childDocs = children.docs;
+    const contents =
+        childDocs.length > 0 ? (
+            childDocs.map((child) => {
+                return (
+                    <li key={(pid || "root") + "child" + child.id}>
```
diboy2: Instead of having to check this condition. Can you just pass in "root" as the pid?

demiankatz: Since pid is guaranteed to be a string (even if an empty one), I think a better solution may be to eliminate this `|| "root"` entirely and instead use a better separator between the parent/child parts. I'll work on that shortly.

client/components/edit/children/ChildList.tsx
```diff
@@ -1,52 +1,43 @@
...
-        start: 0,
-        docs: [],
-    });
+        state: { childStorage },
+        action: { getChildStorageKey, loadChildrenIntoStorage },
+    } = useChildListContext();
     const [page, setPage] = useState<number>(1);
-
+    const key = getChildStorageKey(pid, page, pageSize);
+    const notLoaded = typeof childStorage[key] === "undefined";
```
diboy2: Could you instead do childStorage.hasOwnProperty(key)?

client/components/edit/children/ChildList.tsx
demiankatz: I think [5a94066](https://github.com/FalveyLibraryTechnology/VuDLPrepJS/pull/140/commits/5a94066d422fde516733b18e595187516fd18822) improves this a bit!

client/components/edit/children/ChildList.tsx
demiankatz: Good point. It also is more readable if the variable is called `loaded` instead of `notLoaded`. Revised here: [1d26e5f](https://github.com/FalveyLibraryTechnology/VuDLPrepJS/pull/140/commits/1d26e5fca56b1373ff213a0765f5c4b8c4e6eab6)

client/context/ChildListContext.tsx
demiankatz: (After further discussion outside of this PR, we decided not to pursue this due to mutability concerns; I'm marking this conversation as resolved).

diboy2 (approved):

demiankatz:
I've gotten this to a point of stability and decent test coverage, so I am merging it even though I didn't add all of the needed functionality yet. I'll follow up with additional outstanding work in separate PRs.

PR #142, opened 2022-05-06
Add types to EditorContext.

demiankatz:
This gets most of the missing types added to the EditorContext.

This duplicates some interfaces from the api code -- since api and client are independent code bases, I don't think there's a way to avoid this unless we want to move types into a shared package, which doesn't (yet) seem worth the effort.

I wasn't sure how best to add types to some of the low-level React code, so this is still not 100% perfect, but at least the higher-level stuff is addressed.

### DIFF
--- a/client/context/EditorContext.tsx
+++ b/client/context/EditorContext.tsx
@@ -7,11 +7,42 @@ interface SnackbarState {
     message: string,
     severity: string
 }
+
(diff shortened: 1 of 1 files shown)


diboy2 (approved):

demiankatz:
Thanks, @diboy2!

PR #143, opened 2022-05-12
Update client dependencies; unpin some dev tools.

demiankatz:


demiankatz:
I've been seeing some warnings about incompatibility between my Typescript version and my eslint Typescript plugin. I've unpinned some version numbers here and run updates in an effort to get things more up to date.

demiankatz:
Looks like some upstream changes impacted CSS in a snapshot; I've simply updated snapshots for now, but this is probably a sign that we should do more mocking to reduce fragility of tests.

diboy2:
We can probably modify the config to mock the css.

diboy2 (approved):

PR #144, opened 2022-05-18
Refactor/consolidate editor context management

demiankatz:
I realized upon closer review that some of the work I did for ChildListContext significantly overlapped with EditorContext; this PR merges the two contexts into one and eliminates some redundancies. There is now a single store for all object details, and all components relying on details can draw from the same source (and update accordingly).

This also provided an opportunity to eliminate some "loading" state logic that I wasn't very comfortable with.


crhallberg (approved):
Context work isn't my forte but this does look good to me.

client/context/EditorContext.tsx
```diff
@@ -1,8 +1,22 @@
 import React, { createContext, useContext, useReducer } from "react";
-import { editObjectCatalogUrl, getObjectDetailsUrl } from "../util/routes";
+import { editObjectCatalogUrl, getObjectChildrenUrl, getObjectDetailsUrl } from "../util/routes";
 import { useFetchContext } from "./FetchContext";
 import { extractFirstMetadataValue as utilExtractFirstMetadataValue } from "../util/metadata";
 
+export interface ObjectDetails {
```
diboy2, line 6: Is there a reason to keep object details separate? Can children be add as a key-value to the object details object?

demiankatz: We should probably rename the Children type. The reason for two separate interfaces is that these represent responses from different API endpoints. ObjectDetails describes a single object. The Children interface actually describes a single page from a paginated Solr response, which includes the count of the total result set, the offset of the page, and a docs array containing a page of documents from the overall set. Maybe if we rename this to ChildrenResultPage it would make more sense. What do you think?

demiankatz: I've renamed the interface as proposed above, but happy to rename it again if you have a better suggestion.

diboy2 (approved):

demiankatz:
Thanks, @diboy2!

PR #147, opened 2022-05-26
Add support for editing state and sequences

demiankatz:
This PR adds two significant new editing features to the system.

You can now customize the order of an ordered collection by entering numbers into text boxes in the contents list:

image

You can also see the activation state of every object, for example:

image

You can click a status to change it (and any child objects, if desired):

image

TODO
- [x] Create API for fetching current state/sequence (part of existing object details)
- [x] Create API for updating state
- [x] Create API for updating sequences
- [x] Create UI for displaying state
- [x] Create UI for updating state
  - [x] Basic save functionality
  - [x] "Update children" functionality
  - [x] Improve styles (too much spacing on radio buttons)
  - [x] Simplify "refresh status after save" behavior
- [x] Create UI for displaying sequences
- [x] Create UI for updating sequences
- [x] Add tests related to fetching children / recursive child pids in edit routes
- [x] Add/update tests related to updating state functionality
  - [x] Test API
  - [x] Test React
- [x] Add/update tests related to updating sequence functionality
  - [x] Test API
  - [x] Test React


api/src/routes/edit.test.ts
```diff
@@ -23,7 +24,7 @@ describe("edit", () => {
             allowed_origins: ["http://localhost:3000", "http://localhost:9000"],
         };
         Config.setInstance(new Config(config));
-        pid = "vudl:123";
+        pid = "foo:123";
```
demiankatz, line 29: I changed the pid prefix from vudl to foo here to make it more clear that this is test data.

demiankatz:
@crhallberg / @diboy2, this PR adds two major missing features to the system (see top comment for details/screen shots). Everything is fully functional, but I haven't added test coverage yet. I'm hoping to get a preliminary review to be sure major refactoring is not needed before I lock everything down with tests. Any/all input is welcome!

demiankatz:
@crhallberg / @diboy2, just an update to say that I've made more progress on writing tests. The only things I haven't added tests for are the three new components: ObjectLoader, StateModal and ChildPosition. These seem the most likely to significantly change based on discussion/review. If you have time to review even just one of these, let me know your thoughts and I'll add tests if it seems likely to be nearly stable.

client/components/edit/children/__snapshots__/ChildList.test.tsx.snap
```diff
@@ -152,12 +167,15 @@ exports[`ChildList renders using ajax-loaded root data 1`] = `
...
-      <ul>
+      <ul
+        className="childlist"
+      >
         <li
           key="_child_foo:124"
         >
           <Component
             initialTitle="hello"
+            parentPid=""
```
crhallberg, line 178: Just looking at this line a dozen times in this test makes me wonder if this should be the default value if not included?

crhallberg (approved):
Another super minor thing: the module.css files don't have an empty line at the end of the file and we may as well add them now so that they don't sneak into diffs into the future.

client/components/edit/children/__snapshots__/ChildList.test.tsx.snap
demiankatz: The ChildList defaults pid to blank string... and then it passes down that value as parentPid to all of its children. We could default parentPid to blank string in the Child component, but we'd then have to write more complex logic to omit the attribute when building the ChildList, which I don't think has any real benefit apart from reducing the size of this test snapshot (and has the disadvantage of increasing code complexity).

In any case, I've gone ahead and added the missing whitespace to the CSS files.

client/components/edit/ObjectLoader.tsx
```diff
@@ -0,0 +1,31 @@
...
+        if (!loaded) {
+            loadObjectDetailsIntoStorage(pid);
+        }
+    }, [loaded]);
+    return !loaded ? (
+        <>
+            &nbsp;
+            <CircularProgress size="1em" />
+        </>
+    ) : (
```
diboy2: you can just return null for the else statement.

client/components/edit/StateModal.tsx
```diff
@@ -0,0 +1,188 @@
...
+        const target = getObjectStateUrl(pid);
+        const result = await fetchText(target, { method: "PUT", body: selectedValue });
+        if (result === "ok") {
+            // Clear and reload the cached object, since it has now changed!
+            removeFromObjectDetailsStorage(pid);
+        }
+        return result;
+    };
+
+    const saveChildPage = async (response): Promise<boolean> => {
```
diboy2, line 75: You can repurpose this to a Promise.all. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all

client/components/edit/StateModal.tsx
```diff
@@ -0,0 +1,188 @@
...
+        const expectedTotal = childPidResponse.numFound;
+        let found = 0;
+        let nextResponse = childPidResponse;
+        while (found < expectedTotal) {
+            if (!(await saveChildPage(nextResponse))) {
+                return false;
+            }
+            found += nextResponse.docs.length;
+            if (found < expectedTotal) {
+                const url = getObjectRecursiveChildPidsUrl(details.pid, found, childPageSize);
```
diboy2, line 98: Is it possible to batch all the saves or do these operations in the backend?

client/components/edit/children/ChildPosition.tsx
```diff
@@ -0,0 +1,78 @@
...
+        action: { clearPidFromChildListStorage, removeFromObjectDetailsStorage },
+    } = useEditorContext();
+    const {
+        action: { fetchText },
+    } = useFetchContext();
+    const [statusMessage, setStatusMessage] = useState<string>("");
+
+    // Important: this component assumes it will be called in a context where pid and parent pid
+    // details are already loaded into the context; it will not cause a load to occur by itself.
+    const loaded = Object.prototype.hasOwnProperty.call(objectDetailsStorage, pid);
```
diboy2, line 24: Before calling this component, can you check pid exists before displaying it? Or do you want it to partially render?

client/components/edit/ObjectLoader.tsx
demiankatz: Thanks! Improved here: https://github.com/FalveyLibraryTechnology/VuDLPrepJS/pull/147/commits/f7ecded87ea7bcbf6df3cd2d4ed37c91e36aecdd

diboy2 (approved):

client/context/EditorContext.tsx
```diff
@@ -9,6 +9,7 @@ export interface ObjectDetails {
     models: Array<string>;
```
diboy2, line 9: This file is getting large. Maybe we could create and import modules that concern children states, actions and datastream states, actions into this file. It's something to think about in the future and doesn't have to be necessarily addressed in this pr. I think I also made a mistake of inserting the snackbar actions into the file when that behavior could be useful for all pages of the application.

client/components/edit/children/ChildPosition.tsx
demiankatz: You're right that we can avoid rendering this component in some situations. I've made it conditional with https://github.com/FalveyLibraryTechnology/VuDLPrepJS/pull/147/commits/38b614db254815a5b142966680ee08e805327fa7. I think there's still some value in keeping the loaded checks in the component itself, just in case it gets used in a different context -- this makes it slightly more robust.

client/context/EditorContext.tsx
demiankatz: Yes, agreed. I'll capture a task to refactor this context as a future task.

client/components/edit/StateModal.tsx
demiankatz: I can change it to Promise.all with these diffs:

```
     const saveChildPage = async (response): Promise<boolean> => {
-        for (let i = 0; i < response.docs.length; i++) {
-            const result = await updateStatus(response.docs[i].id);
+        const results = await Promise.all(response.docs.map((doc) => updateStatus(doc.id)));
+        let success = true;
+        results.forEach((result) => {
             if (result !== "ok") {
                 showSnackbarMessage(`Status failed to save; "${result}"`, "error");
                 toggleStateModal();
                 setStatusMessage("");
-                return false;
+                success = false;
             }
-        }
-        return true;
+        });
+        return success;
     };
```

...however, I'm not sure if this is desirable. In the current logic, the work is done serially, which reduces the load on the server and allows clear progress messages. With Promise.all, the work is done in parallel, which potentially puts a heavier load on the server and makes it harder to create coherent messaging.

demiankatz: Were you thinking of an application of Promise.all that retains the existing functionality but makes the code more elegant? Or are these changed characteristics inevitable if we do this refactoring?

client/components/edit/StateModal.tsx
demiankatz: In theory, yes, and this might be a future enhancement. For now, I'm matching the functionality of the legacy system, which does the updates in-browser and provides real-time status updates. This is actually useful in many situations, since the user usually wants to be immediately informed when this task is completed. Some kind of asynchronous batch-editing tool will likely be needed in the future, though, and that could provide an alternative for cases where the user doesn't want to tie up their browser waiting for state changes to finish.

client/components/edit/StateModal.tsx
diboy2: You don't necessarily need to do a forEach after. Promise.all will fail immediately when one of the promises fail. You can throw a particular error at the specific promise that is messing up. https://stackoverflow.com/questions/53981230/using-try-catch-with-promise-all.

demiankatz: The forEach is not to detect failed promises, but rather to validate the API responses. The Promise.all runs a whole bunch of API calls and then returns an array of API responses. It's possible for the API call to fail without causing the promise to fail. I suppose I could put a check inside the updateStatus call to make the promise fail, instead of checking after the fact... but I still think we end up with the same "firehose" effect. If I get 1,000 IDs back, I'd rather work on those serially than try to make my browser simultaneously initiate 1,000 API calls. (And I realize that the browser will probably throttle that in some way... but I feel more comfortable controlling the flow explicitly in this case).

diboy2: Are you trying to show a snackbar message for the first failed result or all failed results? If it's the former, the forEach will continue to iterate through the results even if it encounters a failure. I would use Array.some instead. It will stop at the first failed result and return you the success parameter.

demiankatz: The old logic iterates through the list and stops as soon as something goes wrong -- so it displays a snackbar message for the first failure. If we do things in parallel with Promise.all, we have less control over the flow of things. Array.some is a useful tool, but using it won't change the fact that the promises will all have been resolved before we can check the results. I think at this point I'm more inclined to leave this logic as it is, so we can stop the process if something goes wrong instead of potentially multiplying problems by doing too much work with the system in a bad state. I imagine we could potentially get a similar effect by throwing an exception inside the Promise.all callback, but that seems a bit less controlled and predictable than the current logic.

demiankatz:
I believe that I've gotten test coverage up to adequate levels, so I am merging this now.

demiankatz:
Obviously I'm happy to continue any open discussions and refine logic as needed!

PR #148, opened 2022-05-26
Remove unused/broken Express boilerplate.

demiankatz:
Some of the server.ts setup code is derived from old boilerplate and was causing unhelpful error messages. Removing it makes things behave more predictably (e.g. you actually get 404 errors when you should get 404 errors, instead of a weird stack trace).


diboy2 (approved):

demiankatz:
Thanks, @diboy2!

PR #149, opened 2022-06-22
Added datastream agents editor

diboy2:
image

TODO
- [x] Improve keyboard interface
- [x] Automatically add/submit work in progress when saving (to prevent data loss)

### DIFF
--- a/api/src/models/Config.ts
+++ b/api/src/models/Config.ts
@@ -222,7 +222,19 @@ class Config {
     }
 
     get licenses(): Record<string, License> {
-        return this.ini["licenses"] || {};
+        return this.ini["licenses"] ?? {};
+    }
+
+    get agentDefaults(): Record<string, string> {
+        return this.ini["agent"]["defaults"] ?? {};
+    }
+
+    get agentRoles(): Array<string> {
+        return this.ini["agent"]["roles"] ?? [];
+    }
+
+    get agentTypes(): Array<string> {
+        return this.ini["agent"]["types"] ?? [];
     }
 }
 
--- a/api/src/models/FedoraObject.test.ts
+++ b/api/src/models/FedoraObject.test.ts
@@ -48,6 +48,51 @@ describe("FedoraObject", () => {
         });
     });
 
+    describe("modifyAgents", () => {
+        let agents;
+        let dates;
+        beforeEach(() => {
+            dates = {
+                createDate: "test",
+                modifiedDate: "test",
+            };
+            agents = [
+                {
+                    role: "test1",
+                    type: "test2",
+                    name: "test3",
+                    notes: ["test4"],
+                },
+            ];
+        });
+        test.each([
+            ["zero", [], ""],
+            ["one", ["test4"], "<METS:note>test4</METS:note>"],
+            [
+                "multiple",
+                ["test4", "test5", "test6"],
+                "<METS:note>test4</METS:note><METS:note>test5</METS:note><METS:note>test6</METS:note>",
+            ],
+            ["doubleQuotes", [`""`], "<METS:note>&quot;&quot;</METS:note>"],
+        ])("adds a datastream for an agent with %s notes", async (numString, testInput, notesXml) => {
+            const config = new Config({});
+            const fedora = new Fedora(config);
+            fedoraObject = new FedoraObject(pid, config, fedora, FedoraDataCollector.getInstance());
+            const spy = jest.spyOn(fedoraObject, "addDatastreamFromStringOrBuffer").mockImplementation(jest.fn());
+            agents[0].notes = testInput;
+            const expectedXml =
+                `<METS:agent ROLE="test1" TYPE="test2">` + `<METS:name>test3</METS:name>${notesXml}</METS:agent>`;
+            await fedoraObject.modifyAgents(stream, agents, dates);
+
+            expect(spy).toHaveBeenCalledWith(
+                expect.stringContaining(expectedXml.trim()),
+                stream,
+                "text/xml",
+                [201, 204]
+            );
+        });
+    });
+
     describe("addRelationship", () => {
         it("proxies a call to the fedora service", () => {
             const fedora = Fedora.getInstance();
--- a/api/src/models/FedoraObject.ts
+++ b/api/src/models/FedoraObject.ts
@@ -5,6 +5,7 @@ import Config from "./Config";
 import { DatastreamParameters, Fedora } from "../services/Fedora";
 import FedoraDataCollector from "../services/FedoraDataCollector";
 import { execSync } from "child_process";
+import { Agent } from "../services/interfaces";
 
 export interface ObjectParameters {
     label?: string;
@@ -145,6 +146,29 @@ export class FedoraObject {
         await this.addDatastreamFromStringOrBuffer(licenseXml, stream, "text/xml", [201, 204]);
     }
 
+    async modifyAgents(stream: string, agents: Array<Agent>, agentsAttributes: Record<string, string>): Promise<void> {
+        const { createDate, recordStatus } = agentsAttributes;
+        const agentsXml = agents.reduce((acc, agent) => {
+            const { role, type, name, notes } = agent;
+            const notesXml = notes.reduce((acc, note) => {
+                return acc + `<METS:note>${xmlescape(note)}</METS:note>`;
+            }, "");
+            const agentXml = `<METS:agent ROLE="${xmlescape(role)}" TYPE="${xmlescape(type)}"><METS:name>${xmlescape(
+                name
+            )}</METS:name>${notesXml.trim()}</METS:agent>`.trim();
+            return acc + agentXml;
+        }, "");
+        const documentXml = `
+            <?xml version="1.0" encoding="UTF-8"?>
+            <METS:metsHdr xmlns:METS="http://www.loc.gov/METS/" CREATEDATE="${xmlescape(
+                createDate || new Date().toISOString()
+            )}"${createDate ? ` LASTMODDATE="${xmlescape(new Date().toISOString())}"` : ""}${
+            recordStatus ? ` RECORDSTATUS="${xmlescape(recordStatus)}"` : ""
+        }>${agentsXml}</METS:metsHdr>
+        `.trim();
+        await this.addDatastreamFromStringOrBuffer(documentXml, stream, "text/xml", [201, 204]);
+    }
+
     async addSortRelationship(sort: string): Promise<void> {
         return this.addRelationship("info:fedora/" + this.pid, "http://vudl.org/relationships#sortOn", sort, true);
     }
@@ -186,8 +210,8 @@ export class FedoraObject {
         }
     }
 
-    async getDatastream(datastream: string): Promise<string> {
-        return this.fedora.getDatastreamAsString(this.pid, datastream);
+    async getDatastream(datastream: string, treatMissingAsEmpty = false): Promise<string> {
+        return this.fedora.getDatastreamAsString(this.pid, datastream, treatMissingAsEmpty);
     }
 
     async getDatastreamAsBuffer(datastream: string): Promise<Buffer> {
--- a/api/src/routes/edit.test.ts
+++ b/api/src/routes/edit.test.ts
@@ -123,6 +123,93 @@ describe("edit", () => {
         });
     });
 
+    describe("post /object/:pid/datastream/:stream/agents", () => {
+        let datastreamManager;
+        let agents;
+        beforeEach(() => {
+            datastreamManager = {
+                uploadAgents: jest.fn(),
+            };
+            agents = [
+                {
+                    role: "test1",
+                    type: "test2",
+                    name: "test3",
+                    notes: ["test4", "test5"],
+                },
+            ];
+            jest.spyOn(Database.getInstance(), "confirmToken").mockResolvedValue(true);
+            jest.spyOn(DatastreamManager, "getInstance").mockReturnValue(datastreamManager);
+        });
+
+        it("uploads agents", async () => {
+            datastreamManager.uploadAgents.mockResolvedValue({});
+
+            await request(app)
+                .post(`/edit/object/${pid}/datastream/${datastream}/agents`)
+                .set("Authorization", "Bearer test")
+                .send({ agents })
+                .set("Accept", "application/json")
+                .expect(StatusCodes.OK);
+
+            expect(datastreamManager.uploadAgents).toHaveBeenCalledWith(pid, datastream, agents);
+        });
+
+        it("sends an error status code", async () => {
+            datastreamManager.uploadAgents.mockRejectedValue("upload agents failed");
+
+            await request(app)
+                .post(`/edit/object/${pid}/datastream/${datastream}/agents`)
+                .set("Authorization", "Bearer test")
+                .send({ agents })
+                .set("Accept", "application/json")
+                .expect(StatusCodes.INTERNAL_SERVER_ERROR);
+        });
+    });
+
+    describe("get /object/:pid/datastream/:stream/agents", () => {
+        let datastreamManager;
+        let agents;
+        beforeEach(() => {
+            datastreamManager = {
+                getAgents: jest.fn(),
+            };
+            agents = [
+                {
+                    role: "test1",
+                    type: "test2",
+                    name: "test3",
+                    notes: ["test4"],
+                },
+            ];
+            jest.spyOn(Database.getInstance(), "confirmToken").mockResolvedValue(true);
+            jest.spyOn(DatastreamManager, "getInstance").mockReturnValue(datastreamManager);
+        });
+        afterEach(() => {
+            jest.clearAllMocks();
+        });
+
+        it("gets the agents", async () => {
+            datastreamManager.getAgents.mockResolvedValue(agents);
+
+            const response = await request(app)
+                .get(`/edit/object/${pid}/datastream/${datastream}/agents`)
+                .set("Authorization", "Bearer test")
+                .expect(StatusCodes.OK);
+            expect(response.body).toEqual(agents);
+            expect(datastreamManager.getAgents).toHaveBeenCalledWith(pid, datastream);
+        });
+
+        it("sends an error status code", async () => {
+            datastreamManager.getAgents.mockRejectedValue("get license key fails");
+
+            await request(app)
+                .get(`/edit/object/${pid}/datastream/${datastream}/license`)
+                .set("Authorization", "Bearer test")
+                .expect(StatusCodes.INTERNAL_SERVER_ERROR);
+        });
+    });
+
     describe("get /object/:pid/datastream/:stream/license", () => {
         let datastreamManager;
         let licenseKey;
--- a/api/src/services/MetadataExtractor.ts
+++ b/api/src/services/MetadataExtractor.ts
@@ -1,6 +1,7 @@
 import { DC } from "./Fedora";
 import { DOMParser } from "@xmldom/xmldom";
 import xpath = require("xpath");
+import { Agent } from "./interfaces";
 
 class MetadataExtractor {
     private static instance: MetadataExtractor;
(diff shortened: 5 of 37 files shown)


api/src/models/FedoraObject.test.ts
```diff
@@ -47,6 +47,40 @@ describe("FedoraObject", () => {
...
     });
+    describe("modifyAgents", () => {
+        let agents;
+        beforeEach(() => {
+            agents = [
+                {
+                    role: "test1",
+                    type: "test2",
+                    name: "test3",
+                    notes: ["test4"]
```
demiankatz: To cover all edges, maybe it's worth also having a test with no notes, and a test with multiple notes.

api/src/models/FedoraObject.ts
```diff
@@ -145,6 +145,22 @@ export class FedoraObject {
         await this.addDatastreamFromStringOrBuffer(licenseXml, stream, "text/xml", [201, 204]);
     }
 
+    async modifyAgents(stream: string, agents): Promise<void> {
+        console.log("agents", JSON.stringify(agents));
```
demiankatz: Probably don't want to keep this console.log permanently.

api/src/models/FedoraObject.ts
```diff
@@ -145,6 +145,22 @@ export class FedoraObject {
...
+        const agentsXml = agents.reduce((acc,agent)=> {
+            const { role, type, name, notes }= agent;
+            const notesXml = notes.reduce((acc,note) => {
+                return acc + `<METS:note>${note}</METS:note>`
+            }, "");
+            const agentXml = `<METS:agent ROLE="${role}" TYPE="${type}"><METS:name>${name}</METS:name>${notesXml.trim()}</METS:agent>`.trim();
+            return acc + agentXml;
+        }, "");
+        const documentXml = `
+            <METS:metsHdr xmlns:METS="http://www.loc.gov/METS/" CREATEDATE="2021-08-20T02:28:47.698Z" LASTMODDATE="${new Date().toISOString()}" RECORDSTATUS="">${agentsXml}</METS:metsHdr>
```
demiankatz: Create date should not be hard-coded -- when the agents document is first created, the create date should be set. When it is later modified, the original create date should be retained.

api/src/models/FedoraObject.ts
```diff
@@ -145,6 +145,22 @@ export class FedoraObject {
         await this.addDatastreamFromStringOrBuffer(licenseXml, stream, "text/xml", [201, 204]);
     }
 
+    async modifyAgents(stream: string, agents): Promise<void> {
+        console.log("agents", JSON.stringify(agents));
+        const agentsXml = agents.reduce((acc,agent)=> {
+            const { role, type, name, notes }= agent;
+            const notesXml = notes.reduce((acc,note) => {
+                return acc + `<METS:note>${note}</METS:note>`
```
demiankatz: All values injected into XML should be escaped, to ensure that they do not contain special characters (`<`, `>`, `&`) that could break the markup. You can use the `xmlescape` function for this -- just `import xmlescape = require("xml-escape");`.

api/src/routes/edit.test.ts
```diff
@@ -123,6 +123,32 @@ describe("edit", () => {
...
     });
 
+    describe("post /object/:pid/datastream/:stream/agents", () => {
+        let datastreamManager;
+        let agents;
+        beforeEach(() => {
+            datastreamManager = {
+                uploadAgents: jest.fn(),
+            };
+            agents = [];
```
demiankatz: Would it be better to have some distinctive fake data in here, just to be sure the expectation below isn't met by some other empty array or by a default fallback?

client/hooks/useDatastreamOperation.ts
```diff
@@ -59,6 +60,31 @@ const useDatastreamOperation = () => {
         }
     };
 
+    const uploadAgents = async (agents) => {
+        console.log("agents", agents);
```
demiankatz: Another console.log you can probably remove.

demiankatz (changes requested):
Thanks for the work on this; it's looking good! This review is based on a code-read; I'll let you know if I encounter further issues when I do some hands-on testing a little later this morning.

api/src/models/FedoraObject.ts
demiankatz: It might be good to exercise this with test cases where we send in weird values -- e.g. attribute values with double quotes in them, etc.

api/src/models/Config.ts
```diff
@@ -224,6 +224,14 @@ class Config {
     get licenses(): Record<string, License> {
         return this.ini["licenses"] || {};
     }
+
+    get agentRoles(): Array<string> {
+        return this.ini["agent"]["roles"]
```
demiankatz: Might be good to have these default to something (even if just an empty array) if the configuration is undefined -- I forgot to update vudl.ini at first and everything broke.

demiankatz:
@diboy2, I just did some hands-on testing of this, and I have a few more comments:

1.) We need to be able to configure default values when creating a record for the first time (e.g. the current system automatically populates every record with "DISSEMINATOR / ORGANIZATION / Falvey Memorial Library, Villanova University").

2.) The drop-down menus don't seem to support keyboard shortcuts consistently. Sometimes they work for me, and other times they do not. It would be nice if I could use first letters of options to toggle values while the drop-downs are focused but not expanded.

3.) When adding a note, the input field doesn't clear out; it seems like maybe it should.

4.) If I hit save without explicitly hitting the (+) control, the last row that I entered does not get saved. I understand conceptually why this is the case, but I think users may expect different behavior.

Happy to meet and go over some of this tomorrow, if that's convenient.

api/src/models/FedoraObject.ts
diboy2: Removed

api/src/models/Config.ts
```diff
@@ -222,7 +222,20 @@ class Config {
     }
 
     get licenses(): Record<string, License> {
-        return this.ini["licenses"] || {};
+        return this.ini["licenses"] ?? {};
+    }
+
+    get agentDefaults(): Record<string, string> {
+        console.log(this.ini["agent"]["defaults"])
```
demiankatz: New console.log that needs to be removed (when no longer needed).

client/context/EditorContext.tsx
```diff
@@ -265,8 +287,10 @@ export const useEditorContext = () => {
     const initializeCatalog = async () => {
         try {
             const response = await fetchJSON(editObjectCatalogUrl);
+            console.log("agents catalog: ", response.agents);
```
demiankatz: Another console.log.

api/src/models/FedoraObject.test.ts
```diff
@@ -47,6 +47,60 @@ describe("FedoraObject", () => {
...
+                }
+            ]
+        });
+        
+        it("adds a datastream for an agent with", async () => {
```
demiankatz: With what? Is this test name incomplete? Is this meant to be the "no notes" case? Did something get committed in an incomplete state?

api/vudl.ini.dist
```diff
@@ -333,4 +333,18 @@ name = "Attribution-NonCommercial-NoDerivs"
 uri = "http://creativecommons.org/licenses/by-nc-nd/4.0/"
 [licenses.PUBLIC_DOMAIN]
 name = "CC0 1.0 Universal (Public Domain)"
-uri = "https://creativecommons.org/publicdomain/zero/1.0/"
\ No newline at end of file
+uri = "https://creativecommons.org/publicdomain/zero/1.0/"
+
+# Agent
+[agent]
+
```
demiankatz, line 340: Can we put an example in here of how to set agent defaults? Even if it's commented out, it would probably be helpful for clarity.

api/src/models/FedoraObject.ts
```diff
@@ -145,6 +145,21 @@ export class FedoraObject {
...
+        const agentsXml = agents.reduce((acc,agent)=> {
+            const { role, type, name, notes }= agent;
+            const notesXml = notes.reduce((acc,note) => {
+                return acc + `<METS:note>${xmlescape(note)}</METS:note>`
+            }, "");
+            const agentXml = `<METS:agent ROLE="${xmlescape(role)}" TYPE="${xmlescape(type)}"><METS:name>${name}</METS:name>${notesXml.trim()}</METS:agent>`.trim();
+            return acc + agentXml;
+        }, "");
+        const documentXml = `
+            <METS:metsHdr xmlns:METS="http://www.loc.gov/METS/" CREATEDATE="${new Date().toISOString()}" LASTMODDATE="${new Date().toISOString()}" RECORDSTATUS="">${agentsXml}</METS:metsHdr>
```
demiankatz: This is now going to change the create date every time the record is modified. We need to establish the create date when the datastream is first created, and then never change it when other parts of the stream are modified. I think the simplest approach may be to actually pass the original create date as part of the post when the agents are modified. Obviously this would lend itself to spoofing/manipulation (as opposed to reading the value from the existing datastream before rewriting it, which is another possible solution) but it seems simpler and faster, and I can't think of a reason why anyone would want to falsify this data, or why it would matter.

api/src/models/FedoraObject.ts
demiankatz: Thanks for fixing the problem -- I'm going to leave this thread open for now as a reminder to add test cases when time permits. I can help with that if you like!

demiankatz (changes requested):
Thanks for the progress on this. I've left a few new comments on the latest changes, and we can discuss further at our meeting tomorrow!

demiankatz:
@diboy2, in addition to the above review, I've added a couple of TODO boxes to the top-level comment so we don't lose track of open items from my [earlier comment](https://github.com/FalveyLibraryTechnology/VuDLPrepJS/pull/149#issuecomment-1163417370), which is partially addressed as of this writing.

api/src/models/Config.ts
diboy2: Made change.

api/src/models/FedoraObject.test.ts
diboy2: Sorry. I was planning to implement a foreach

api/src/models/FedoraObject.ts
diboy2: I'm still thinking about the workflow since I have to extract the create date and keep it as part of the xml content. I might plan to extract the create date from the front end and check if it exists.

diboy2: I decided to remove the security issue, by keeping the extraction of the create date in the backend.

demiankatz:
@diboy2, thanks for the progress on this. Is this currently meant to be in a working state? I just tried to test it, and I'm getting an "Internal server error" when I try to save the AGENTS datastream. Are you seeing the same thing?

I don't currently have a test record with an existing AGENTS datastream. Is it possible there is a problem where some of the new code fails if you are creating AGENTS for the first time, but editing existing records works? (Just a wild guess).

demiankatz:
Thanks for the latest progress, @diboy2. A few comments (some of which we discussed in chat yesterday, but which I want to document here for future reference):

1.) The XML document should be updated to start with the standard XML header (`<?xml version="1.0" encoding="UTF-8"?>`).

2.) If there is an existing RECORDSTATUS attribute in a document, we should persist that forward to the new document the same we we do the CREATEDATE.

3.) I think we should only include the default values from the configuration file if there are no existing entries in the agents list -- so an empty document gets populated with defaults, and a non-empty document gets populated using an empty form. The defaults are only useful the first time around. I think maybe the best way to approach this is that the default doesn't control what goes into the "add row" form, but rather is treated as the existing contents of the document, if there are not yet any existing contents. This approach would also make it easier to implement multiple rows of default values, if we needed to do that in the future.

4.) While I understand your concerns about the possibility of scrolling, I think it probably makes sense to move the "add" controls to the bottom of the list, especially now that things are saving automatically. That would make the form feel a lot more WYSIWYG than it currently does. Realistically, I've never seen an agent document with more than 5-6 rows, so I think scrolling is only a problem in very extreme cases.

5.) I've seen some strange behavior with the save buttons. If I start to add a row and then change my mind, there doesn't seem to be a straightforward way to get out of the situation, because I can't delete my partially-formed row, and I can't save until I finish adding that row. The obvious workaround is to add the row with fake data and then immediately remove it, but that's not very natural. One possible solution, which I think works best if we move the edit row to the bottom as discussed in #4 above, would be to change the behavior so that, if the contents of the row are not valid, the `(+)` control becomes `(-)` and allows the row-in-progress to be removed (which would then turn the last preceding row into the new edit row with a `(+)` control). You would be allowed to save the form in any state except completely empty, but you would be warned if any of your rows were partially empty/incomplete. An alternative thought: maybe the `(+)` control isn't useful at all, and we're better off with an "Add Row" button at the top, and always having a `(-)` control to remove rows.

Anyway, in spite of all my comments above, I think this is very close to being in a "good enough" state, and I'm sure you want to get this finished as much as I do. I'm not trying to turn this into a protracted process, and if any of my suggestions sound unhelpful or seem likely to take a very long time to implement, I'm very open to compromise! :-)

diboy2:
@demiankatz I've addressed 1,2,3, and possibly 4. Could we meet monday to discuss 5.? I'm not sure I understand what the expected behavior should be.

api/src/models/FedoraObject.ts
```diff
@@ -158,11 +158,12 @@ export class FedoraObject {
...
         }, "");
         const documentXml = `
+            <?xml version="1.0" encoding="UTF-8"?>
             <METS:metsHdr xmlns:METS="http://www.loc.gov/METS/" CREATEDATE="${xmlescape(
                 createDate || new Date().toISOString()
-            )}"${
-            createDate ? ` LASTMODDATE="${new Date().toISOString()}"` : ""
-        } RECORDSTATUS="">${agentsXml}</METS:metsHdr>
+            )}"${createDate ? ` LASTMODDATE="${new Date().toISOString()}"` : ""}${
+            recordStatus ? ` RECORDSTATUS="${recordStatus}"` : ""
```
demiankatz: We should probably xmlescape the record status to be on the safe side (unless it's being passed through in pre-escaped form). I think it's unlikely this will ever contain special characters, but better safe than sorry.

demiankatz:
@diboy2, thanks for the progress on this; it's very close to where it needs to be. I'm just seeing some strange behavior when I edit existing data -- I often end up with double blank rows on the end of the list. I also wonder if putting a red warning saying "this row is incomplete and will not be saved" on the last row would make more sense than blocking out the "save and close" button. But we can discuss that (and related ideas) later today when we meet.

api/src/models/FedoraObject.ts
diboy2: mde change.

api/src/services/MetadataExtractor.ts
```diff
@@ -165,6 +165,73 @@ class MetadataExtractor {
...
+                if (attr.nodeName == "ROLE") {
+                    agent.role = attr.nodeValue;
+                }
+                if (attr.nodeName == "TYPE") {
+                    agent.type = attr.nodeValue;
+                }
+            });
+            Object.values(relation.childNodes).forEach((childNode) => {
+                if (childNode.nodeName == "METS:name") {
+                    agent.name = childNode.firstChild.nodeValue;
```
demiankatz: I think it might be a good idea to check for the presence of firstChild before assuming it has a nodeValue. When there are empty values in the XML (which _shouldn't_ happen, but clearly _can_ happen due to bugs), I'm getting fatal `"Cannot read property 'nodeValue' of null` messages from the /agents API endpoint which prevent me from opening the "edit agents" modal, and I suspect it's related to this bit of code.

api/vudl.ini.dist
```diff
@@ -333,4 +333,23 @@ name = "Attribution-NonCommercial-NoDerivs"
...
+roles[] = "DISSEMINATOR"
+roles[] = "CUSTODIAN"
+roles[] = "IPOWNER"
+roles[] = "CREATOR"
+
+types[] = "ORGANIZATION"
+types[] = "INDIVIDUAL"
+
+[agent.defaults]
+name = "Falvey Memorial Library, Villanova University"
```
demiankatz: We should change this example to something like "My Library, My University" just to keep it generic and consistent with the rest of the example configuration. (I thought that I actually made the edit here previously, but I think it may have gotten force-pushed out of existence).

client/components/edit/datastream/DatastreamAgentContent.tsx
```diff
@@ -0,0 +1,146 @@
...
+    useEffect(() => {
+        const callGetAgents = async () => {
+            const currentAgents = await getAgents();
+            if (!currentAgents.length) {
+                currentAgents.push({ role, type, name, notes: [] });
+                setHasChanges(true);
+            }
+            setCurrentAgents(currentAgents);
+        };
+        callGetAgents();
```
demiankatz: Should we have some kind of loading status indicator mechanism here? Since the getAgents call can take a while, it might prevent confusion if the controls were not visible until the data was fully loaded.

client/components/edit/datastream/DatastreamAgentContent.tsx
```diff
@@ -0,0 +1,146 @@
...
+                                }}
+                            >
+                                Save Changes
+                            </Button>
+                        )}
+                        {hasChanges && (
+                            <Button
+                                disabled={!canSave}
+                                onClick={() => {
+                                    uploadAgents([...currentAgents, addAgent]);
```
demiankatz: Why are you including addAgent here, but not in the previous uploadAgents call? At this point, shouldn't both save buttons work the same, apart from one closing the dialog and the other not? Could this be related to the empty name bug you mentioned yesterday?

Also: might it be worth including logic in the uploadAgents function to simply filter out rows with no name in them? Hopefully the UI design should prevent bad values from slipping through, but an extra filter layer will ensure that bad data is never sent to the server.

demiankatz (changes requested):
Thanks for the ongoing progress, @diboy2. I ran into the "empty row" bug you mentioned yesterday, and I took a closer look at the code and left some comments that might help you fix it. Please let me know if you need/want me to take a deeper look at any point. Happy to help further, but I suspect this may be enough of a nudge in the right direction. :-)

client/components/edit/datastream/DatastreamAgentContent.tsx
diboy2: Sure.

client/components/edit/datastream/DatastreamAgentContent.tsx
diboy2: I think you might have discovered the actual bug.

api/src/services/MetadataExtractor.ts
diboy2: Made change.

api/vudl.ini.dist

client/components/edit/datastream/DatastreamAgentContent.tsx
diboy2: Made change.

client/components/edit/datastream/DatastreamAgentContent.tsx
diboy2: Removed it. the addAgent was empty which caused the additional row to be saved.

client/context/EditorContext.tsx
diboy2: Removed.

api/src/models/FedoraObject.ts
diboy2: Added double quotes scenario

client/components/edit/datastream/DatastreamAgentContent.tsx
```diff
@@ -120,7 +132,9 @@ const DatastreamAgentContent = (): React.ReactElement => {
                             <Button
                                 disabled={!canSave}
                                 onClick={() => {
-                                    uploadAgents([...currentAgents, addAgent]);
+                                    setIsLoading(true);
+                                    uploadAgents(currentAgents);
```
demiankatz: Do we need an await here to prevent the loading indicator from clearing before the load is done?

demiankatz (changes requested):
Thanks for the fixes and improvements! I have just one question about the new code based on a quick review. I ran out of time before I could do hands-on testing today, but I'll take a closer look tomorrow.

demiankatz:
@diboy2, I think the latest changes aren't exactly what we want. Now, if I fill out the agents form and save it, the bottom row does not get saved unless I explicitly hit "(+)" an extra time. Also, if I edit the agents and type in the bottom row, I can't save the form until after I hit "(+)".  Is it possible that, rather than removing the `addAgent` part of `uploadAgents([...currentAgents, addAgent]);`, we actually want to only include the `addAgent` part when it is non-empty?

client/components/edit/datastream/DatastreamAgentContent.tsx
diboy2: Made change.

api/src/routes/edit.test.ts
diboy2: Made change.

demiankatz (approved):
Thanks, @diboy2, I think this is working very well now. Please feel free to merge if you're feeling ready.

client/components/edit/datastream/DatastreamAgentsContent.test.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+        const wrapper = mount(<DatastreamAgentsContent />);
+        await act(async () => {
+            wrapper.find("button.agentsCancelButton").simulate("click");
+            wrapper.update();
+        });
+
+        expect(datastreamOperationValues.getAgents).toHaveBeenCalled();
+        expect(editorValues.action.setCurrentAgents).toHaveBeenCalled();
+        expect(editorValues.action.toggleDatastreamModal).toHaveBeenCalled();
```
demiankatz, line 125: Would it make sense to assert that uploadAgents is not called in this context, just to be safe?

demiankatz (changes requested):
Thanks for the progress, @diboy2 -- I had one small (and entirely optional) suggestion. The only other issue is that the build is failing in GitHub; it looks like maybe some snapshots have gotten out of sync somehow. Let me know if you want me to spend any time investigating that!

client/components/edit/datastream/DatastreamAgentsContent.test.tsx
diboy2: Made change.

demiankatz:
Thanks for the ongoing work, @diboy2. I noticed that the build was failing because of out-of-sync snapshots, so I just updated them and pushed them up. I think we're all set to merge now!

demiankatz (approved):

PR #150, opened 2022-07-21
Add missing test for DatastreamMetadataModalContent.

demiankatz:
@diboy2, while testing, I just happened to notice that one of the datastream content components was missing a test. Fortunately, adding the test was mostly a copy-and-paste operation, so I think this should be pretty trivial to approve. :-)


diboy2 (approved):

demiankatz:
Thanks, @diboy2!

PR #152, opened 2022-08-10
Edit parent relationships

demiankatz:
This PR adds "edit parent relationship" functionality to the editor.

TODO
- [x] Wire up "Edit Parents" modal
- [x] Display existing parent details
- [x] Create "PID picker" utility
  - [x] Allow direct entry via keyboard
  - [x] Allow selection via tree
  - [x] Add "set to last" button to Position control
  - [x] Allow "favorite list" via configuration
- [x] Implement "add new parent" functionality
  - [x] Back end API
  - [x] Front end UI
- [x] Implement "delete parent" functionality
  - [x] Back end API
  - [x] Front end UI
- [x] Add test coverage for everything
  - [x] Back end tests
  - [x] Front end tests


api/src/services/Fedora.ts
```diff
@@ -310,6 +310,42 @@ export class Fedora {
...
+    /**
+     * This method removes an isMemberOf relationship from a pid/parent pid pair.
+     *
+     * @param pid       PID to update
+     * @param parentPid Parent PID to update
+     */
+    async deleteParentRelationship(pid: string, parentPid: string): Promise<void> {
+        const predicate = "info:fedora/fedora-system:def/relations-external#isMemberOf";
+        const targetPath = "/" + pid;
+        const insertClause = "";
```
diboy2, line 329: Are these empty strings for a reason? Can you just add the empty string into the parameters?

client/components/edit/Breadcrumbs.tsx
```diff
@@ -1,130 +1,36 @@
...
-                childLookups: {},
-                records: {},
-            };
-            const url = getObjectParentsUrl(pid);
-            try {
-                data = processBreadcrumbData(await fetchJSON(url));
-            } catch (e) {
-                console.error("Problem fetching breadcrumb data from " + url);
-            }
-            setTreeData(data);
```
diboy2, line 86: This is a pretty good refactor. We usually want to abstract out the transformation logic and just leave the render logic.

api/src/services/Fedora.ts
demiankatz: When we delete a relationship, we intentionally do not insert anything -- hence the empty insertClause string. I thought it might be more readable to put this in a named variable rather than directly in the function call, but it would work equally well either way. Happy to change if you think this way is overkill.

client/components/edit/EditParentsButton.tsx
```diff
@@ -0,0 +1,29 @@
...
+export interface ObjectStatusProps {
+    pid: string;
+}
+
+export const EditParentsButton = ({ pid }: ObjectStatusProps): React.ReactElement => {
+    const {
+        state: { objectDetailsStorage },
+        action: { setParentsModalActivePid, toggleParentsModal },
+    } = useEditorContext();
+    const loaded = Object.prototype.hasOwnProperty.call(objectDetailsStorage, pid);
```
diboy2: What's the purpose of this function? Can you just set a default value?

client/components/edit/children/SelectableChild.tsx
```diff
@@ -0,0 +1,44 @@
...
+    selectCallback: (pid: string) => void;
+    initialTitle: string;
+}
+
+export const SelectableChild = ({ pid, selectCallback, initialTitle }: SelectableChildProps): React.ReactElement => {
+    const {
+        state: { objectDetailsStorage },
+    } = useEditorContext();
+    const [expanded, setExpanded] = useState<boolean>(false);
+    const loaded = Object.prototype.hasOwnProperty.call(objectDetailsStorage, pid);
```
diboy2, line 19: Is there a way to filter out the undefined pids before rendering components so you don't have to do this additional check?

client/components/edit/EditParentsButton.tsx
demiankatz: The `<ObjectLoader>` component has a side effect of loading object details into storage... so this makes the component self-contained. You can put it anywhere, and it will load the data it needs to operate, without depending on some other higher-level component. Whether or not this is a good idea is a separate question, but that's the purpose of the load check.

If we want to keep this approach, we could probably simplify the function to `return loaded ? <button>...</button> : <ObjectLoader ... />` instead of what it is currently doing, and it would have the same effect.

Alternatively, we could eliminate all the ObjectLoader stuff from here, and do the checking at a higher-level component where the button is displayed.

client/components/edit/children/SelectableChild.tsx
demiankatz: I don't think so. When we render the list of children, we have PIDs and titles from the Solr index, but these might be out of date. Wiring each SelectableChild up to the objectDetailsStorage ensures that we display the most up-to-date state of those objects (and redraw them if other operations on the page cause changes to Fedora). It feels to me like trying to account for this at a higher level would be more complicated and might lead to larger re-renders than are necessary.

client/util/Breadcrumbs.tsx
```diff
@@ -0,0 +1,87 @@
...
+ * Analyze the raw breadcrumb data, identifying top-level nodes (i.e. nodes with
+ * no parents) and creating lookup tables for children. This makes it easier to
+ * render breadcrumbs from left to right and to find all relevant trails.
+ */
+export function processBreadcrumbData(data: TreeNode): TreeData {
+    const queue: Array<TreeNode> = [data];
+    const topNodes: Array<string> = [];
+    const childLookups: Record<string, Array<string>> = {};
+    const records: Record<string, TreeNode> = {};
+    while (queue.length > 0) {
```
diboy2: breadth first search?

client/util/Breadcrumbs.tsx
```diff
@@ -0,0 +1,87 @@
...
+            queue.push(parent);
+            if (typeof childLookups[parent.pid] === "undefined") {
+                childLookups[parent.pid] = [current.pid];
+            } else {
+                childLookups[parent.pid].push(current.pid);
+            }
+        });
+    }
+    return {
+        topNodes: Array.from(new Set(topNodes)), // deduplicate
```
diboy2: You can maybe instantiate topNodes into Set first.

client/util/Breadcrumbs.tsx
diboy2: You can maybe implement depth first search instead since array.shift() is actally O(n) for javascript instead of O(1).

demiankatz: Good to know. This code is a direct port of an algorithm that I didn't write, so I'll need to spend some time analyzing it. That's probably a separate task, though, since this logic is already present in dev and has just been relocated for this PR.

client/util/Breadcrumbs.tsx
demiankatz: So you're just suggesting that I define topNodes as a Set instead of an Array so there's no need to convert back and forth? If that's the case, it makes sense to me... but again, I will probably address that as a separate PR since it doesn't directly relate to the work in this one.

client/components/edit/EditParentsButton.tsx
demiankatz: In closer review of the code, there's no compelling reason to incorporate ObjectLoader functionality into the EditParentsButton, so I've simplified it with commit b5f4eaf7c3ca05720e11f1d967be1c506ba67405. Thanks for catching that bit of over-engineering!

api/src/models/FedoraObject.ts
```diff
@@ -123,6 +123,14 @@ export class FedoraObject {
         );
     }
 
+    async addParentRelationship(parentPid: string): Promise<void> {
+        return this.addRelationship(
+            "info:fedora/" + this.pid,
```
diboy2, line 128: "info:fedora/" is the same for all parameters. You can maybe add this to part of the base url.

api/src/routes/edit.ts
```diff
@@ -319,6 +341,72 @@ edit.put("/object/:pid/state", requireToken, pidSanitizer, bodyParser.text(), as
...
 
 const pidAndParentPidSanitizer = sanitizeParameters({ pid: pidSanitizeRegEx, parentPid: pidSanitizeRegEx });
+edit.put(
+    "/object/:pid/parent/:parentPid",
+    requireToken,
+    pidAndParentPidSanitizer,
+    bodyParser.text(),
+    async function (req, res) {
+        try {
+            const pid = req.params.pid;
```
diboy2: You can do.` const { pid, parentPid: parent} = req.params;`

api/src/models/FedoraObject.ts
demiankatz: In this instance, we're setting up an RDF triple where all three parts happen to use info:fedora/ URIs. However, this is not universally guaranteed. It wouldn't make sense to hard-code info:fedora into the generic addRelationship method, and I don't see an obvious way to simplify what we have here. I think the redundancy may be unavoidable.

api/src/routes/edit.ts
```diff
@@ -319,6 +341,72 @@ edit.put("/object/:pid/state", requireToken, pidSanitizer, bodyParser.text(), as
...
+            const parent = req.params.parentPid;
+            const pos = parseInt(req.body);
+
+            // Validate the input
+            if (pid == parent) {
+                res.status(400).send("Object cannot be its own parent.");
+                return;
+            }
+            const parentData = await FedoraDataCollector.getInstance().getHierarchy(parent);
+            if (parentData.getAllParents().includes(pid)) {
```
diboy2, line 360: You can maybe throw here instead and setup your errors in the catch statement instead.

api/src/routes/edit.ts
```diff
@@ -319,6 +341,72 @@ edit.put("/object/:pid/state", requireToken, pidSanitizer, bodyParser.text(), as
...
+);
+edit.delete("/object/:pid/parent/:parentPid", requireToken, pidAndParentPidSanitizer, async function (req, res) {
+    try {
+        const pid = req.params.pid;
+        const parent = req.params.parentPid;
+        const fedora = Fedora.getInstance();
+
+        // Validate the input
+        const fedoraData = await FedoraDataCollector.getInstance().getHierarchy(pid);
+        const legalParent = fedoraData.parents.reduce((previous, current) => {
```
diboy2: This might be Array.some(). You're only looking for one of them to be a parent, right?

api/src/services/FedoraCatalog.ts
```diff
@@ -70,6 +75,30 @@ class FedoraCatalog {
...
         }, {});
     }
+
+    async getFavoritePids(): Promise<Record<string, string>> {
+        const pids = this.config.favoritePids;
+        const result = {};
+        if (pids.length > 0) {
+            const queryParts = pids.map((pid) => {
+                return `id:"${pid.replace('"', "")}"`;
+            });
```
diboy2: Can't you add the join statement here and make it query?

api/src/services/FedoraCatalog.ts
```diff
@@ -70,6 +75,30 @@ class FedoraCatalog {
...
+            const query = queryParts.join(" OR ");
+            const solrResponse = await this.solr.query(this.config.solrCore, query, {
+                fl: "id,title",
+                rows: queryParts.length.toString(),
```
diboy2: Isn't this pid.length.toString?

api/src/routes/edit.ts
demiankatz: Hooray for destructuring! I've optimized things a bit in b8e764d653501fc8bb90239b58a568ea79d673c1.

api/src/routes/edit.ts
demiankatz: I tried to refactor this but I couldn't find a way to do it that didn't end up with worse-looking code. The challenge is that we want to return a 500 status if there's an unexpected technical problem, but a 400 status if there's an input validation error. In order to do that with a single try/catch, we would need to have a way of flagging the type of the error so the catch clause could set the correct status. There are a variety of ways I can think of to do that, but they all feel kind of hacky and actually add lines of code. Can you think of a technique that I'm overlooking?

api/src/routes/edit.ts
demiankatz: Good point, though actually I think it's Array.find() since we want to return the matching element rather than just a boolean status; see 8369602bf43f504c2d28902a82cd4256480c4a2d.

api/src/services/FedoraCatalog.ts
demiankatz: Good point -- see d51f3ea7b23d613bca5c7a5b3eb7bb925463450d.

api/src/services/FedoraCatalog.ts
demiankatz: Also addressed by d51f3ea7b23d613bca5c7a5b3eb7bb925463450d.

api/src/routes/edit.ts
diboy2: ok

client/util/Breadcrumbs.tsx
demiankatz: Fixed in #165.

PR #165, opened 2022-09-02
Breadcrumb fixes and improvements

demiankatz:
TODO
- [x] Simplify set handling
- [x] Expand test coverage
- [x] Optimize algorithm
- [x] Fix bug: breadcrumb displays sometimes include duplicate paths


demiankatz:
@diboy2, following up on your feedback from #152, this PR greatly improves the breadcrumb processing. Perhaps most importantly, it fixes a bug where we were using an Array instead of a Set, which was causing duplicate breadcrumb trails to appear under some circumstances. It also adds 100% test coverage of all breadcrumb-related code, and reduces our logic from two searches to a single search thanks to a major optimization proposed and developed by @crhallberg.

diboy2 (approved):

PR #167, opened 2022-10-18
Use BlurSavingTextField for more responsive Agent editing.

demiankatz:
Editing Agent text fields could be a bit laggy due to the validation calculations taking place during typing. This PR replaces the TextField with a BlurSavingTextField to improve responsiveness. It also expands BlurSavingTextField to support arbitrary extra attributes on the underlying TextField. The only downside is that error state doesn't go away until you move away from the input field, but this seems like a price worth paying to avoid lagginess while typing.


diboy2 (approved):

demiankatz:
Thanks, @diboy2!

PR #174, opened 2022-11-30
Improve breadcrumbs

demiankatz:
This PR provides breadcrumb navigation back to the main menu (and intermediate screens) from every part of the application. It is mostly achieved through a consistent shared component, with the exception of the job paginator, where the overall screen layout follows a different pattern. (We might make this more consistent in future, but for now this seems like an incremental improvement at least).


diboy2:
You should consider sharing the snackbar component.

demiankatz:
@diboy2, making the snackbar more generic is already on the to-do list. Another thing I need to do is move the to-do list into GitHub issues here so it's more visible. :-)

diboy2 (approved):

demiankatz:
Thanks, @diboy2!

PR #175, opened 2022-12-09
Support job-specific queues.

demiankatz:
This PR makes it possible to send different job types to different queues for more granular worker management.

TODO
- [x] Adjust the worker startup script to allow specification of a queue name
- [x] Test everything (including formal unit tests)


demiankatz:
Once this is working, these pages may be helpful for systemd setup:

https://unix.stackexchange.com/questions/288236/have-systemd-spawn-n-processes

https://unix.stackexchange.com/questions/212438/how-to-stop-all-units-belonging-to-the-same-target

PR #201, opened 2022-12-23
Add configuration to set Tika config file.

demiankatz:
I've noticed an issue where Tika performs a really expensive Tesseract operation when reading Microsoft Word documents during indexing. I can't figure out how to disable it; I thought perhaps by providing a custom XML configuration [as described on the Tika wiki](https://cwiki.apache.org/confluence/display/TIKA/TikaOCR), I could change the behavior. Unfortunately, it doesn't seem to help. However, since I went to the trouble of building a mechanism to provide a custom XML configuration in order to test this, I thought this work might be worth saving.

TODO:
- [x] Test that this really has an effect (e.g. find a configuration setting that has a visible impact on output, and make sure providing a config file has the expected effect)
- [x] Write unit test coverage
- [x] Only merge if there's a meaningful use case for the functionality


demiankatz:
I have confirmed a meaningful use case. Disabling OCR on PDF files requires additional settings. This is the configuration I ended up with:

```
<?xml version="1.0" encoding="UTF-8"?>
<properties>
  <parsers>
    <parser class="org.apache.tika.parser.DefaultParser">
      <parser-exclude class="org.apache.tika.parser.ocr.TesseractOCRParser"/>
    </parser>
    <parser class="org.apache.tika.parser.ocr.TesseractOCRParser">
      <params>
        <param name="skipOcr" type="bool">true</param>
      </params>
    </parser>
    <parser class="org.apache.tika.parser.pdf.PDFParser">
      <params>
        <param name="ocrStrategy" type="string">NO_OCR</param>
      </params>
    </parser>
  </parsers>
</properties>
```

I confirmed that pointing to an XML file containing these contents bypasses the OCR, and omitting the config results in very slow performance.

It's possible that there's an additional similar setting needed for Microsoft Office documents, but I have not yet researched that.

PR #256, opened 2023-02-08
Create Global Context

### LINKED ISSUE #185: Refactor snackbar to application level
A "snackbar" component was added for displaying status messages from the datastream component. It may be useful to move this to a higher level for easier reuse by other components.

crhallberg:
- [x] Snackbar
- [x] Convert isXModalOpen to function
- [x] Ensure completeness of functionality
- [x] Write tests
- [x] Investigate test warning: "A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks. Active timers can also cause this, ensure that .unref() was called on them." [CANNOT REPRODUCE]
- [x] Close #185 when merging


crhallberg:
@demiankatz I've moved the ThemeMenu component and associated global state to the dark mode PR (#258).

client/components/edit/ObjectStatus.tsx
```diff
@@ -18,7 +22,7 @@ export const ObjectStatus = ({ pid }: ObjectStatusProps): React.ReactElement =>
     const stateTxt = details.state ?? "Unknown";
     const clickAction = () => {
         setStateModalActivePid(pid);
-        toggleStateModal();
+        toggleModal("state");
```
demiankatz: You can remove the import of toggleStateModal up on line 17 above.

client/pages/_app.tsx
```diff
@@ -22,4 +22,4 @@ function MyApp({ Component, pageProps }: { Component: React.ReactNode }): React.
         </PaginatorContextProvider>
     );
 }
-export default MyApp;
+export default MyApp;
```
demiankatz: You seem to have lost the end-of-file line break from this file for some reason.

client/hooks/useDatastreamOperation.ts
```diff
@@ -51,14 +55,14 @@ const useDatastreamOperation = () => {
                 message: text,
                 severity: "success",
             });
-            toggleDatastreamModal();
+            toggleModal();
```
demiankatz: Don't all the calls to toggleModal in this file need a parameter (`"datastream"`)?

client/hooks/useDatastreamOperation.test.ts
```diff
@@ -33,6 +40,11 @@ describe("useDatastreamOperation", () => {
                 },
             },
         };
+        globalValues = {
+            action: {
+                setSnackbarState: jest.fn(),
```
demiankatz, line 45: You should also remove setSnackbarState from editorValues below.

client/hooks/useDatastreamOperation.test.ts
```diff
@@ -1,6 +1,12 @@
 import { beforeEach, describe, expect, it, jest } from "@jest/globals";
 import useDatastreamOperation from "./useDatastreamOperation";
 
+const mockUseGlobalwContext = jest.fn();
```
demiankatz: Typo: `GlobalwContext`

client/context/GlobalContext.tsx
```diff
@@ -0,0 +1,133 @@
...
+};
+
+/**
+ * Update the shared states of react components.
+ */
+const globalReducer = (state: GlobalState, { type, payload }: { type: string, payload: unknown}) => {
+    if (type == "OPEN_MODAL") {
+        return {
+            ...state,
+            isModalOpen: { [payload]: true }
```
demiankatz: I think if you have multiple modal states in play, this approach will lose all but the most recently changed modal. You might have to do something more like:

```
const isModalOpen = {
    ...state.isModalOpen,
    [payload]: true,
};
return {
    ...state,
    isModalOpen,
};
```

...and that's enough code that maybe it should be refactored to a getStateWithChangedModalSetting function, so you can share the logic between OPEN_MODAL and CLOSE_MODAL.

client/context/GlobalContext.tsx
```diff
@@ -0,0 +1,133 @@
...
+    const [state, dispatch] = useReducer(globalReducer, globalContextParams);
+    const value = { state, dispatch };
+    return <GlobalContext.Provider value={value}>{children}</GlobalContext.Provider>;
+};
+
+export const useGlobalContext = () => {
+    const {
+        state: {
+            // Modal control
+            isModalOpen,
```
demiankatz: I had originally envisioned exposing an `isModalOpen` function rather than the raw data, to make it easier to change the internal implementation if we wanted to. (I had actually imagined an array of strings representing open modals, rather than a map of modal names to Booleans). However, I don't really see anything wrong with this approach -- just highlights the fact that there are always multiple ways to do a thing. :-)

client/context/GlobalContext.test.tsx
```diff
@@ -0,0 +1,31 @@
+import { beforeEach, describe, expect, it, jest } from "@jest/globals";
+import { renderHook, act } from "@testing-library/react-hooks";
+import { GlobalContextProvider, useGlobalContext } from "./GlobalContext";
+
+describe("useGlobalContext", () => {
+    describe("setSnackbarState", () => {
```
demiankatz, line 6: We should also add some modal-related tests.

client/context/EditorContext.tsx
```diff
@@ -219,6 +207,11 @@ export const EditorContextProvider = ({ children }) => {
 };
 
 export const useEditorContext = () => {
+    const {
+        action: {
+            toggleModal
+        }
+    } = useGlobalContext();
```
demiankatz: Do we really need to import this here? I think it might be better to just leave it separate (it doesn't seem to be used anyway).

client/context/EditorContext.test.tsx
```diff
@@ -235,33 +235,6 @@ describe("useEditorContext", () => {
         });
     });
 
-    describe("setSnackbarState", () => {
```
demiankatz, line 238: We should delete the various toggle*Modal tests too.

client/components/edit/parents/ParentsModal.tsx
```diff
@@ -5,16 +5,25 @@ import DialogTitle from "@mui/material/DialogTitle";
...
 import ObjectLoader from "../ObjectLoader";
 import ParentList from "./ParentList";
 import ParentPicker from "./ParentPicker";
 
 const ParentsModal = (): React.ReactElement => {
     const {
-        state: { isParentsModalOpen, objectDetailsStorage, parentsModalActivePid },
-        action: { toggleParentsModal },
+        state: { isModalOpen },
+        action: { openModal, closeModal },
```
demiankatz: openModal is not used.

client/components/edit/datastream/DatastreamModal.tsx
```diff
@@ -23,15 +24,15 @@ const DatastreamModalContent = ({ datastreamModalState }: { datastreamModalState
 
 const DatastreamModal = (): React.ReactElement => {
     const {
-        state: { datastreamModalState, isDatastreamModalOpen },
-        action: { toggleDatastreamModal },
-    } = useEditorContext();
+        state: { datastreamModalState },
+        action: { isModalOpen, openModal, closeModal },
```
demiankatz: isModalOpen is currently a state, not an action.

demiankatz (changes requested):
Thanks, @crhallberg, this is definitely moving in the right direction, but there are some bits that don't seem to be fully fleshed out yet (you probably already know that :-) ). Hopefully the observations below will help to move things along. I'd also do a global search for "toggle" in case there are other modal-related calls that haven't been done yet. I was expecting to see more than I did while reviewing, but maybe I just overestimated what to expect. :-)

client/pages/_app.tsx
crhallberg: When I reverted it, the original didn't have it. I'll add it back in.

client/components/edit/ObjectStatus.tsx
crhallberg: Done.

client/context/GlobalContext.tsx
crhallberg: I refactored it as a function and it worked. I think this is the nicer form of the interface, so I'm going to commit that.

client/context/GlobalContext.tsx
```diff
@@ -93,7 +100,7 @@ export const useGlobalContext = () => {
         });
     };
     const toggleModal = (modal: string) => {
-        if (isModalOpen[modal]) {
+        if (modalOpenStates[modal]) {
```
demiankatz: Might want a `?? false` here for safety.

demiankatz (changes requested):
@crhallberg, just one small suggestion here based on recent changes. Beyond that, what's the state of this project? I see we still have failing tests. How can I help? Do you want me to help work on any automated or hands-on testing, or should I just give you more time to continue this? :-)

client/components/edit/ObjectStatus.tsx
```diff
@@ -8,17 +9,20 @@ export interface ObjectStatusProps {
...
+        action: { setStateModalActivePid },
     } = useEditorContext();
     const loaded = Object.prototype.hasOwnProperty.call(objectDetailsStorage, pid);
     const details = loaded ? objectDetailsStorage[pid] : {};
 
     const stateTxt = details.state ?? "Unknown";
     const clickAction = () => {
         setStateModalActivePid(pid);
-        toggleStateModal();
+        toggleModal("state");
```
demiankatz: Should this be openModal for consistency with other components, or is there a reason to use toggleModal?

client/components/edit/StateModal.test.tsx
```diff
@@ -23,34 +30,46 @@ jest.mock("../../context/FetchContext", () => ({
...
         fetchContextValues = {
             action: {
                 fetchJSON: jest.fn(),
                 fetchText: jest.fn(),
             },
         };
+        mockUseGlobalContext.mockReturnValue(globalValues);
+        mockUseEditorContext.mockReturnValue(editorValues);
         mockUseFetchContext.mockReturnValue(fetchContextValues);
+        globalValues.action.isModalOpen.mockReturnValue(true);
```
demiankatz, line 82: We should probably add `expect(globalValues.action.isModalOpen).toHaveBeenCalledWith("state");` to the tests in this suite. I did that for the other tests where I mocked isModalOpen yesterday, but I neglected to do it here.

client/components/edit/datastream/DatastreamModal.tsx
```diff
@@ -23,15 +24,17 @@ const DatastreamModalContent = ({ datastreamModalState }: { datastreamModalState
...
+        action: { closeModal, isModalOpen },
+    } = useGlobalContext();
 
     return (
         <Dialog
             className="datastreamModal"
-            open={isDatastreamModalOpen}
-            onClose={toggleDatastreamModal}
+            open={isModalOpen("datastream")}
+            onClose={() => closeModal("datastream")}
```
demiankatz, line 37: In other components, we define a local `closeXModal` function instead of using the `() => ...` syntax. Both approaches are valid, but I wonder if we should choose one or the other for consistency. I don't feel a lot of pressure to change anything, but let me know if you have an opinion!

client/context/GlobalContext.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+const reducerMapping: Record<string, string> = {
+    // Snackbar
+    SET_SNACKBAR_STATE: "snackbarState",
+};
+
+/**
+ * Update the shared states of react components.
+ */
+const globalReducer = (state: GlobalState, { type, payload }: { type: string, payload: unknown}) => {
```
demiankatz: I wonder if payload should be `SnackbarState | unknown` to match what it was back in EditorContext. Not sure how helpful that typehint really is, but I guess it's a start that we can work on making more specific over time.

demiankatz (changes requested):
@crhallberg, I've done another manual review of this code following yesterday's work. I think this is in very good shape, but I had a few minor suggestions and discussion points. I'm happy to handle some or all of the work, depending on your preferences. Once all of this is sorted out, I'll do some hands-on testing to assure myself it's all good, but I don't expect any problems that should block a merge.

One other general comment: I did carefully inspect all the snapshots generated by the snapshot tests, and there were no unexpected/unwanted changes there. The trick to making those diffs readable is to add `?w=1` to the GitHub URL to ignore whitespace.

client/components/edit/StateModal.test.tsx
demiankatz: I just took care of this.

client/context/GlobalContext.tsx
crhallberg: What do you think of this type construction?

```ts
type Action =
    | { type: 'OPEN_MODAL'; payload: string }
    | { type: 'CLOSE_MODAL'; payload: string }
    | { type: 'SET_SNACKBAR_STATE'; payload: SnackbarState };

/**
 * Reducer to manage global state updates.
 */
const globalReducer = (state: GlobalState, action: Action): GlobalState => {
```

demiankatz: If that works and covers all the currently supported scenarios, I think that sounds great!

client/context/GlobalContext.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+    const setSnackbarState = (snackbarState: SnackbarState) => {
+        dispatch({
+            type: "SET_SNACKBAR_STATE",
+            payload: snackbarState
+        });
+    };
+
+    return {
+        state,
```
demiankatz: The original code didn't return the whole state because it was intentionally hiding parts of it to make it "private" and encourage use of the action methods instead. I'm not sure if significant harm is done by changing that, but just an FYI.

demiankatz (changes requested):
@crhallberg, thanks for the progress here! Just one new comment (which may not require any action). Please also see my two existing comments from May, 2023 and let me know if you want to do anything with those. I suspect we can leave the reported inconsistency alone, but I think it may make sense to replace the "toggle" with "open" to be more specific. Once all this is sorted, I think we can merge!

client/components/edit/datastream/DatastreamModal.tsx
demiankatz: I think we can leave this alone for now.

client/components/edit/ObjectStatus.tsx
demiankatz: I addressed this in [f48b273](https://github.com/FalveyLibraryTechnology/VuDL/pull/256/commits/f48b273b4389299e6da8d7c39c3e2b29a1b77cec).

demiankatz (approved):
I believe that our work here is done!

PR #258, opened 2023-02-08
Improve Styles, including Dark Mode

### LINKED ISSUE #255: Dark mode
The current UI for the software is very light -- it would be nice to have a toggle to dark mode for users who prefer a darker interface.

crhallberg:
Based off of the [Global Context PR](https://github.com/FalveyLibraryTechnology/VuDLPrepJS/pull/256).

- [ ] Map grays to palette
- [ ] Implement dark theme
- [ ] Touch up padding and margins
- [ ] Add test coverage to new code
- [ ] Close #255 when merging


demiankatz (changes requested):
@crhallberg, now that #256 is merged, there are a ton of conflicts here (since this work was based on a very early version of #256). Can you resolve the conflicts (or, if it's easier, close this PR and port the relevant changes over to a new one)?

client/components/ThemeMenu.tsx
```diff
@@ -0,0 +1,35 @@
+import React from "react";
+import InputLabel from '@mui/material/InputLabel';
+import MenuItem from '@mui/material/MenuItem';
+import FormControl from '@mui/material/FormControl';
+import Select, { SelectChangeEvent } from '@mui/material/Select';
```
demiankatz, line 5: You're importing SelectChangeEvent but not using it.

client/components/ThemeMenu.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+import { useGlobalContext } from "../context/GlobalContext";
+
+export default function ThemeMenu() {
```
demiankatz, line 9: We should add a test for this component once it is finalized and before we merge this PR.

client/components/edit/ObjectButtonBar.tsx
```diff
@@ -19,7 +19,7 @@ const ObjectButtonBar = ({ pid }: ObjectButtonBarProps): React.ReactElement => {
             <ObjectStatus pid={pid} />
             <EditParentsButton pid={pid} />
             <button onClick={() => clearPidFromChildListStorage(pid)}>
-                <Refresh style={{ height: "14px" }} titleAccess="Refresh children" />
+                <Refresh style={{ height: "20px", verticalAlign: "sub" }} titleAccess="Refresh children" />
```
demiankatz, line 23: Should we give this a class and move the styling to CSS? I'm pretty sure the direct style here was just my laziness, but if we're going to get fancier, maybe we should do things "the right way."

client/styles/globals.css
```diff
@@ -1,16 +1,134 @@
...
+    --thumbnail-bg: #dde1e6;
+    --thumbnail-label: #121619;
+    --thumbnail-focus-bg: #82cfff;
+    --thumbnail-number-bg: #ffffff;
+    --thumbnail-number-text: #4d5358;
+    --thumbnail-preview-border: #000000;
+
+    /* Editor */
+
+    --object-status-active: #bef264;
```
demiankatz, line 60: I think the status colors need to be more clearly defined -- the active/inactive need to be distinct from each other so you can tell them apart at a glance, but these new colors make them look too similar:

image

Is there a reason we can't keep the standard red/green/yellow?

client/components/ThemeMenu.tsx
```diff
@@ -0,0 +1,35 @@
...
+    } = useGlobalContext();
+
+    function changeTheme(e) {
+        setUserTheme(e.target.value);
+    }
+
+    return (
+        <FormControl className="user-theme__menu">
+            <InputLabel id="user-theme__label">Theme</InputLabel>
+            <Select
```
demiankatz, line 22: When it comes time to test this, take a look at the test for my bulk editor -- it's a bit tricky to test MUI select controls, but that test does it successfully. (You have to trigger a MouseUp action, if I remember correctly...)

client/context/GlobalContext.tsx
```diff
@@ -149,3 +175,50 @@ export default {
...
     useGlobalContext
 }
+
+/* User Theme */
+
+// Get system theme from CSS media queries
+function systemTheme(): ThemeOption {
+    let defaultTheme = "light" as ThemeOption;
+
+    if (typeof window != "undefined" && typeof window.matchMedia != "undefined") {
```
demiankatz, line 185: watchMedia is not defined in server-side code, so I had to add the second part of this check to prevent test failures. We can mock the method if we want to actually test this logic (and we should, once it's ready).

client/context/GlobalContext.tsx
```diff
@@ -149,3 +175,50 @@ export default {
...
+function localLoadUserTheme(): ThemeOption {
+    if (typeof window != "undefined") {
+        let mediaTheme = (localStorage.getItem("vudl-theme") ?? "system") as ThemeOption;
+
+        applyUserThemeToBody(mediaTheme);
+
+        return mediaTheme;
+    }
+
+    return "system" as ThemeOption;
```
demiankatz, line 223: Can you return `ThemeOption.system` instead of `"system" as ThemeOption`? If that works, it seems easier to understand. (And we should use that pattern for all the theme references).

client/pages/_app.tsx
```diff
@@ -5,11 +5,15 @@ import { PaginatorContextProvider } from "../context/PaginatorContext";
...
-import "../styles/client.css";
 import "../styles/justgrid.css";
+import "../styles/globals.css";
 import "@fortawesome/fontawesome-free/css/all.min.css";
+
+import "../styles/application.css";
+import "../styles/client.css";
+
 import LogoutButton from "../components/LogoutButton";
+import ThemeMenu from "../components/ThemeMenu";
```
demiankatz, line 16: You import the ThemeMenu, but you don't use it. Indeed, I don't see a theme menu anywhere in the application at this point, though I do notice that the "log out" button has moved from right-aligned to left-aligned for some reason.

client/styles/justgrid.css
```diff
@@ -1 +1,20 @@
-.row:after,.row:before{clear:both;content:"";display:table}.col,.row{box-sizing:border-box;width:100%}.col{float:left}.col.one{width:8.333%}.col.two{width:16.666%}.col.three{width:25%}.col.four{width:33.333%}.col.five{width:41.666%}.col.six{width:50%}.col.seven{width:58.333%}.col.eight{width:66.666%}.col.nine{width:75%}.col.ten{width:83.333%}.col.eleven{width:91.666%}@media(max-width:600px){.row:not(.static)>.col{width:100%}}
\ No newline at end of file
+/*
```
demiankatz, line 1: Can/should we remove this commented-out code? Or should we consider revising the JobPaginator to use the MUI grid so we don't need this file at all? (I'm pretty sure this is only being used in JobPaginator.tsx, though maybe there are some other references I'm missing.

demiankatz (changes requested):
Thanks for the progress made here, @crhallberg! There were a few small problems that were causing test failures, so I went ahead and fixed them. This seems to be functional now, though I couldn't find a way to actually try out dark mode. See below for more detailed comments and suggestions.

demiankatz:
I also merged a couple of small fixes from here into dev (whitespace adjustments, removing the stray semicolon in JobPaginator) to reduce the number of files in this PR.

demiankatz (changes requested):
Now that #337 is merged, this needs a refresh.

Do you want to resolve conflicts here and get this cleaned up, or should we just close it and start again from a new foundation?

PR #277, opened 2023-10-16
Update dependencies (October, 2023).

demiankatz:


demiankatz:
@crhallberg, just trying to do some routine maintenance here, and things are weirdly breaking on GitHub (though tests are passing and things are generally working in my test VM). Any ideas what might be going on?

Also note that there's a new major version of Next.js out; we should probably work through that upgrade when time permits so as not to accumulate too much technical debt!

PR #281, opened 2023-10-23
Move API-specific dependencies out of top-level package.json.

demiankatz:
This PR is designed to improve consistency in our handling of tools. Some of the tools used only in the api/ folder were imported through the top-level package.json. It makes more sense to move these to where they are actually used, and to keep the top-level package.json focused only on scripts that cross both parts of the project.

While moving things around, I also noticed some naming inconsistencies and missing documentation, so I addressed those details while I was at it.


crhallberg (approved):
Nodemon can be moved to API when we remove the WSL support commands.

PR #299, opened 2023-11-19
Use Solr cache to reduce Fedora and Redis load.

demiankatz:
TODO
- [x] Retrieve data from Solr cache instead of Fedora when possible
- [x] Invalidate cache when reindex jobs are queued
- [x] Use lock files in cache to avoid duplicate index jobs
- [x] Test coverage for everything

### DIFF
--- a/api/src/jobs/Index.test.ts
+++ b/api/src/jobs/Index.test.ts
@@ -1,6 +1,7 @@
 import Index from "./Index";
 import { Job } from "bullmq";
 import { NeedleResponse } from "../services/interfaces";
+import SolrCache from "../services/SolrCache";
 import SolrIndexer from "../services/SolrIndexer";
 
 jest.mock("../services/SolrIndexer");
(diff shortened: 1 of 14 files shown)


crhallberg:
The only case I'm worried about is if the server restarts but the queue remains. Should we clear out lock files? Check them against the queue?

demiankatz:
> The only case I'm worried about is if the server restarts but the queue remains. Should we clear out lock files? Check them against the queue?

@crhallberg, I don't think this is actually a problem, because the queue is stored persistently in Redis. If the server restarts, the queue won't go away; it will just pick up where it left off.

I do agree that it would be useful to monitor for old lock files, since this would be a good way to detect potential bugs and problems (we could just do a "find" command looking for *.lock files older than a day or two -- this would be a good candidate for a Nagios task). However, I don't think we should actually find any under normal circumstances.

demiankatz:
I've added some smarter cache invalidation and ensured test coverage for everything; I think this is ready to go!

PR #341, opened 2025-09-08
make leading underscores in ingest consistent

### LINKED ISSUE #206: Some jobs are ingested with a leading underscore in the filename
Sometimes the auto-generated name for newly-ingested objects starts with an underscore, and sometimes it does not. This seems to be due to an inconsistency in whether or not the job directory path has a trailing slash on it (possibly a difference between manually triggering a job build, and using the auto-script). This should be normalized so that job names are the same regardless of trailing slashes in their paths.

Jason-Benson:
TODO
- [ ] Resolve #206 when merging.


demiankatz (changes requested):
Thanks, @Jason-Benson. Looks like this change has caused a number of tests to fail. In some cases, this may be a sign that we need to change the tests; in other cases, it may be revealing some bad assumptions in other code that are exposed by normalizing the directory path here. Next time we get together, I should give you a tour of the test suite so you can work through fixing these things!

api/src/models/Job.ts
```diff
@@ -16,7 +16,7 @@ class Job {
     queue: QueueManager;
 
     constructor(dir: string, config: Config, queue: QueueManager) {
-        this.dir = dir;
+        this.dir = dir.endsWith("/") ? dir : dir + "/";
```
demiankatz, line 19: It also occurred to me that if we wanted to trim the trailing slash rather than normalize it on, we could use a regular expression, like:

```
this.dir = dir.replace(/\/+$/, "");
```

... while I still think I _like_ having the leading underscore, I do wonder if it's more technically correct not to have it. Maybe the right thing to do would be to fix the bug by stripping the trailing slash, and then add a feature where we can define a "job name prefix" value in the config file. That might be best of both worlds. I also suggest this in part because fewer tests may fail if we go in the opposite direction.

Anyway, just giving options for you to think about. We can discuss further!

demiankatz:
@Jason-Benson, since it seems like a good time to revive this PR, here's my proposed path forward:

1.) As per my [comment thread above](https://github.com/FalveyLibraryTechnology/VuDL/pull/341#discussion_r2331532742), I think it would make sense to revise this so that we strip slashes off of paths instead of always adding slashes on. This will allow you to revert most of the changes in this PR and just focus on the assignment of the `this.dir` property. The end result will be that job names will no longer start with an underscore under any circumstances.

2.) Because I _like_ my jobs to start with underscores, I think we should add a config setting to set a job name prefix. Maybe we could add something like:

```
# Prefix to add to all newly-ingested job names (e.g. "_" to sort new jobs to top of holding area)
job_name_prefix = "_"
```

...and then in [Config.ts](https://github.com/FalveyLibraryTechnology/VuDL/blob/dev/api/src/models/Config.ts) we can add a getter method to retrieve that setting, and then we can use it directly as part of job name construction in the job here.

Does that make any sense?

PR #343, opened 2025-12-10
Depend1

Jason-Benson:
This is my first round of dependency updates.


api/src/models/FedoraObject.test.ts
```diff
@@ -21,7 +21,7 @@ describe("FedoraObject", () => {
         filename = "test1";
         stream = "test2";
         mimeType = "test3";
-        jest.spyOn(fs, "readFileSync").mockReturnValue(buffer);
+        (fs.readFileSync as jest.Mock).mockReturnValue(buffer);
```
demiankatz: I'm not sure I understand this one, though it clearly works. I tried reverting the change just to understand what was happening -- and I found that we could also make the error go away by using a `string` instead of a `Buffer` as the type of `buffer`. (I just populated it with fake junk instead of using `Buffer.alloc()`. I don't fully understand why this is happening, so I'm not sure which solution is better. Maybe you've done better research than me and have a stronger idea. In any case, we should discuss before merging -- but this is the only change in the PR that raised any concern for me, and everything is passing on my end.

demiankatz (commented):
Thanks, @Jason-Benson! See below for one area we might want to discuss -- but this looks great!

api/src/models/FedoraObject.test.ts
demiankatz: (Root cause seems to be the type definitions for the `fs` module using a deprecated `NonSharedBuffer` type -- but I have no idea why that would be or what it means; I don't know much about how the `@types/node` library is maintained, but you certainly seem to be using an appropriately up-to-date version).

demiankatz (approved):
Thanks, @Jason-Benson, this is looking good so far!

PR #344, opened 2025-12-12
Depend2

Jason-Benson:
Dependency updates for:
body-parser
bull-arena
bullmq


api/src/services/Authentication.ts
```diff
@@ -71,7 +71,7 @@ class Authentication {
             }.bind(this),
             // TODO: implement a logout function here:
             () => null,
-        );
+    )) as unknown as passport.Strategy;
```
demiankatz: This looks like a red flag to me. If we have to cast saml.Strategy to passport.Strategy, it suggests that there may be incompatible libraries in play together. Unfortunately, SAML is used for our single sign-on, so it's hard to test outside of our real environment... thus, an incompatibility there wouldn't become obvious before we deployed this to our test server.

The key is that saml is being pulled in from the `@node-saml/passport-saml` package. You might want to check if there's a newer version of that available, and whether upgrading it eliminates the need for this typecasting magic.

demiankatz (changes requested):
This will be easier to parse through after #343 is merged, but in the meantime, I did spot one thing that might require a little more work.

api/src/models/FedoraObject.test.ts
```diff
@@ -21,7 +21,7 @@ describe("FedoraObject", () => {
         filename = "test1";
         stream = "test2";
         mimeType = "test3";
-        jest.spyOn(fs, "readFileSync").mockReturnValue(buffer);
+        (fs.readFileSync as jest.Mock).mockReturnValue(buffer);
```
demiankatz: Looks like the merging didn't undo this unnecessary change. I'll just force it through a review suggestion:

```suggestion
        jest.spyOn(fs, "readFileSync").mockReturnValue(buffer);
```

demiankatz (changes requested):
Just one suggestion, which I'll go ahead and apply...

demiankatz (approved):
Looks good to me -- thanks, @Jason-Benson!

PR #345, opened 2025-12-15
Depend3b

Jason-Benson:
dependency updates that don't require code changes


demiankatz (changes requested):
Thanks, @Jason-Benson! I did some searching of the code and hands-on testing of these dependencies to reassure myself that everything is safe, and I ran into a couple things of note:

1.) connect-session-knex and http-errors are not used anywhere in the code, so I went ahead and removed them from this branch. (No further action needed on this point).

2.) formidable is used for processing uploads -- and it appears that the new major version breaks something, as uploads no longer work. We should either roll that one back or else figure out how to fix it. It looks like the tests didn't catch this one because we mock out the formidable implementation, so the upstream changes did not break the test. We'll likely have to fix the test at the same time as the other code.

api/src/routes/edit.ts
```diff
@@ -197,8 +197,13 @@ function uploadFile(req, res, next) {
             return;
         }
         try {
+            let fileEntry = files?.file ?? files?.[Object.keys(files || {})[0]];
+            if (Array.isArray(fileEntry)) {
+                fileEntry = fileEntry[0];
+            }
+            const filepath = fileEntry.filepath ?? fileEntry.path ?? fileEntry.filePath;
+            const mimetype = fileEntry.mimetype ?? fileEntry.type ?? "";
```
demiankatz: Is this logic attempting to be back-compatible with the older version of Formidable? If so, I don't think that's necessary -- since we're locking in the newer version, I think it would be safe to assume that's what we're going to use. If that's the case, could we simplify to something like:


```suggestion
            let fileEntry = files[0];
            const filepath = fileEntry.filepath;
            const mimetype = fileEntry.mimetype;
```

(That's just a guess -- I am probably getting some structural details wrong... but I think there has to be an easier way than what we have here, since it seems to be accounting for many theoretical possibilities that may not really exist. I'd just add `console.log(files)` to the code to see what the structure looks like, and then adjust the logic to accommodate that).

demiankatz (changes requested):
Thanks, @Jason-Benson -- I suspect this can be made simpler. See below! Happy to get together and look at this if it helps (though probably not before Friday, since I'm in meetings for the rest of today and out of office tomorrow).

api/src/routes/edit.ts
Jason-Benson: That's pretty much what I had before, but it tripped up the test. I'll switch it back and look into correcting the test instead.

demiankatz: Yeah, the test is probably sending in fake data in the old format, so if you change the test's simulated data to match the new reality, then everything will be happy.

demiankatz (approved):
Thanks, @Jason-Benson. It looks like you trusted me too much -- the proposed change I suggested didn't actually work. I just did a little hands-on testing and found my mistake, which I have corrected in commit [e2cbd28](https://github.com/FalveyLibraryTechnology/VuDL/pull/345/commits/e2cbd281653e4bb01288f3019b3a06963f91444d). I'll merge this as soon as GitHub Actions pass!

PR #346, opened 2025-12-18
Depend4

Jason-Benson:
More dependency updates


demiankatz (approved):
Thanks, @Jason-Benson, this looks good to me. I did some tweaking to MetadataExtractor to simplify the changes -- rather than using typecasting in multiple places, I created a centralized wrapper function to make the types of the XPath extractors more predictable.

PR #347, opened 2026-01-05
Upgrade to Express 5 (along with some other minor dependency updates)

Jason-Benson:
Updating the @types/express dependency to the current version required making changes to authentication.ts
getSamlStrategy had mixed verify-signature forms, so TypeScript picked an overload that caused the library to pass different arguments at runtime — the done parameter ended up being a non-function, producing the "done is not callable" error. Making both callbacks the same form, giving explicit types, and removing the invalid placeholder ensured that overload resolution, runtime argument order, and TypeScript types all lined up.


api/tsconfig.json
```diff
@@ -3,7 +3,8 @@
       "sourceMap": true,
       "allowJs": true,
       "target": "es5",
-      "outDir": "./dist"
+      "outDir": "./dist",
+      "skipLibCheck": true
```
demiankatz, line 7: Is this `skipLibCheck` setting really needed, or was it only necessary as a transitional step? The [documentation](https://www.typescriptlang.org/tsconfig/#skipLibCheck) suggests that it may be needed in certain circumstances but might be best avoided when possible. I can live with it if we have to, but if we can do without, that would be better -- and if it is needed, maybe we should open an issue to remind ourselves to work it out in the future.

api/src/services/Authentication.ts
```diff
@@ -95,7 +99,7 @@ class Authentication {
             passport.use(this.getLocalStrategy());
         } else if (authStrategy === "saml") {
             const samlStrategy = this.getSamlStrategy();
-            passport.use(samlStrategy);
+            passport.use(samlStrategy as unknown as passport.Strategy);
```
demiankatz:
```suggestion
            passport.use(samlStrategy);
```

Do we actually need this change? samlStrategy should always be a subclass of passport.Strategy, so I don't think there's a need to recast it. When I remove this, tests and lint still pass for me, and the build still builds.

api/src/services/Authentication.ts
```diff
@@ -60,17 +61,20 @@ class Authentication {
...
+                let user: User | null = null;
+                if (this.isLegalUsername(profile?.nameID)) {
                     const db = Database.getInstance();
                     user = await db.getOrCreateUser(profile.nameID);
                 }
-                done(null, user);
-            }.bind(this),
-            // TODO: implement a logout function here:
-            () => null,
+                done(null, user ? (user as unknown as Record<string, unknown>) : undefined);
```
demiankatz: I think I prefer using `null` here instead of `undefined` since we're explicitly saying "there is no valid user" rather than "we haven't defined a user yet." But maybe I'm splitting hairs. If nothing else, it's shorter. ;-)

```suggestion
                done(null, user ? (user as unknown as Record<string, unknown>) : null);
```

api/tsconfig.json
demiankatz: Update: while I had the code up, I tried turning this off to see what would happen -- and the answer is a few `Private identifiers are only available when targeting ECMAScript 2015 and higher.` errors. I tried changing the `target` setting from `es5` to `es6` to address that concern, and that unlocked hundreds of different errors. Perhaps you have already been down this path too. I suspect that in the long term, switching to `es6` (which is now a pretty widespread baseline standard) would be another good modernization step. But in the short term, this is obviously another rabbit hole... so I'd suggest we keep this as-is and open an issue as a reminder to review/modernize the Typescript config in the future.

demiankatz (changes requested):
Thanks, @Jason-Benson! See below for a few thoughts and suggestions.

api/src/services/Authentication.ts
Jason-Benson: Argument of type 'import("/opt/vudl/api/node_modules/@node-saml/passport-saml/lib/strategy").Strategy' is not assignable to parameter of type 'import("/opt/vudl/api/node_modules/@types/passport/index").Strategy'.
  Types of property 'authenticate' are incompatible.
    Type '(req: Request<ParamsDictionary, any, any, ParsedQs, Record<string, any>>, options: AuthenticateOptions) => void' is not assignable to type '(this: StrategyCreated<Strategy, Strategy & StrategyCreatedStatic>, req: Request<ParamsDictionary, any, any, ParsedQs, Record<string, any>>, options?: any) => any'.
      Types of parameters 'req' and 'req' are incompatible.
        Property 'param' is missing in type 'import("/opt/vudl/api/node_modules/@types/express/index").Request<import("/opt/vudl/api/node_modules/@types/express-serve-static-core/index").ParamsDictionary, any, any, qs.ParsedQs, Record<string, any>>' but required in type 'import("/opt/vudl/api/node_modules/@node-saml/passport-saml/node_modules/@types/express/index").Request<import("/opt/vudl/api/node_modules/@node-saml/passport-saml/node_modules/@types/express-serve-static-core/index").ParamsDictionary, any, any, qs.ParsedQs, Record<...>>'.ts(2345)

demiankatz:
Superseded by #350.

PR #352, opened 2026-01-20
2gb upload limit fix

Jason-Benson:
To get around node's 2gb limit on handling, I reworked the ingest and metadata handling to stream the files in chunks.


api/src/jobs/Metadata.ts
```diff
@@ -22,16 +21,18 @@ class MetadataProcessor {
     }
 
     async addMasterMetadataDatastream(): Promise<void> {
+        console.log(`Adding master metadata datastream to ${this.pid}`);
```
demiankatz: Do we still need this console.log?

api/src/jobs/Metadata.ts
```diff
@@ -22,16 +21,18 @@ class MetadataProcessor {
...
         const fedoraObject: FedoraObject = FedoraObject.build(this.pid, null, this.config);
-        const dataStream: Buffer = await fedoraObject.getDatastreamAsBuffer("MASTER");
-        const contentFile = tmp.fileSync();
-        fs.writeFileSync(contentFile.name, dataStream);
-        await fedoraObject.addMasterMetadataDatastream(contentFile.name);
-        fs.truncateSync(contentFile.name, 0);
-        fs.rmSync(contentFile.name);
+        console.log("FedoraObject.build: Done");
```
demiankatz: How about this one?

api/src/models/FedoraObject.ts
```diff
@@ -65,6 +66,7 @@ export class FedoraObject {
             params.logMessage ?? "Adding datastream " + id + " to " + this.pid + " with " + data.length + " bytes",
         );
         await this.fedora.addDatastream(this.pid, id, params, data, expectedStatus);
+        console.log(`Added datastream ${id} to ${this.pid}`);
```
demiankatz: Is it safe to remove all the console.log messages from this file as well? If you want to keep any of these messages in the long term, it would be better in this file to call `this.log()` which will send the message to the job processing log instead of to the console.

api/src/models/FedoraObject.ts
```diff
@@ -221,11 +256,16 @@ export class FedoraObject {
         return this.fedora.getDatastreamAsBuffer(this.pid, datastream);
     }
 
+    async getDatastreamToTempFile(datastream: string, treatMissingAsEmpty = false): Promise<string> {
```
demiankatz: Any particular reason the method is called `getDatastreamToTempFile` here but `downloadDatastreamToTempFile` in the Fedora class? I'd be inclined to use the "download" name in both places for consistency, though you can persuade me otherwise if you kept them separate for a specific reason. :-)

api/src/services/Fedora.ts
```diff
@@ -70,7 +72,12 @@ export class Fedora {
             password: this.config.fedoraPassword,
         };
         const options = Object.assign({}, auth, _options);
-        return http(method, url, data, options);
+
+        return http(method, url, data, options).catch((err) => {
+            console.error(`Request failed for ${method.toUpperCase()} ${url}:`, err);
+            console.error("Full error:", JSON.stringify(err, Object.getOwnPropertyNames(err), 2));
```
demiankatz: Is JSON.stringify needed here? I thought that console.error should be able to handle a raw exception object appropriately... though maybe this makes it less verbose?

api/src/services/Fedora.ts
```diff
@@ -95,14 +102,18 @@ export class Fedora {
     }
 
     async getDatastreamAsBuffer(pid: string, datastream: string, treatMissingAsEmpty = false): Promise<Buffer> {
+        console.log("getDatastreamAsBuffer:Start");
```
demiankatz: I imagine you may also be able to remove console.log calls from this file now that you're done.

api/src/models/FedoraObject.ts
```diff
@@ -77,7 +79,26 @@ export class FedoraObject {
...
+        await new Promise<void>((resolve, reject) => {
+            const rs = fs.createReadStream(filename);
+            rs.on("data", (chunk: Buffer) => {
+                md5Hash.update(chunk);
+            });
+            rs.on("end", () => resolve());
+            rs.on("error", (err) => reject(err));
+        });
+        const md5 = md5Hash.digest("hex");
+        const digestHeader = `md5=${md5}`;
```
demiankatz: By default, we used to  provide md5 and sha512 hashes, but you're only precomputing the md5. Is that intentional or an oversight?

api/src/services/Fedora.ts
```diff
@@ -264,18 +341,30 @@ export class Fedora {
...
-            Digest: "md5=" + md5 + ", sha-512=" + sha,
         };
+
+        // If caller supplied a precomputed digest (for streaming upload), use it.
+        if (precomputedDigest && precomputedDigest.length > 0) {
+            headers.Digest = precomputedDigest;
+        } else {
+            // For string/Buffer payloads, compute digests here.
+            if (typeof data === "string" || Buffer.isBuffer(data)) {
+                const md5 = crypto.createHash("md5").update(data).digest("hex");
```
demiankatz, line 333: We seem to have lost the sha512 hash here as well.

api/src/services/Fedora.ts
```diff
@@ -201,6 +215,69 @@ export class Fedora {
...
+    async downloadDatastreamToTempFile(pid: string, datastream: string, treatMissingAsEmpty = false): Promise<string> {
+        const path = pid + "/" + datastream;
+        const urlPath = path[0] == "/" ? path.slice(1) : path;
+        const url = this.config.restBaseUrl + "/" + urlPath;
+
+        const auth = {
+            username: this.config.fedoraUsername,
+            password: this.config.fedoraPassword,
+        };
+        const options = Object.assign({}, auth);
```
demiankatz: We use Object.assign in other parts of the code to combine multiple objects together, but since there is only one object here (auth), I think this call is unnecessary. You could just assign the values directly to an options variable and skip the assign step.

api/src/services/Fedora.ts
```diff
@@ -201,6 +215,69 @@ export class Fedora {
...
+     * Download a datastream directly to a temporary file to avoid buffering
+     * large files into memory.
+     *
+     * @param pid Record id
+     * @param datastream Which stream to request
+     * @param treatMissingAsEmpty If true, return empty temp file on 404
+     */
+    async downloadDatastreamToTempFile(pid: string, datastream: string, treatMissingAsEmpty = false): Promise<string> {
+        const path = pid + "/" + datastream;
+        const urlPath = path[0] == "/" ? path.slice(1) : path;
```
demiankatz: I don't think this slash-trimming step is necessary here. Looks like you've based this code on the generic _request method, but since a pid will never start with a slash, I don't think you need to mirror all the logic of the more generic method.

api/src/services/TikaExtractor.ts
```diff
@@ -34,6 +34,17 @@ class TikaExtractor {
         fs.rmSync(filename); // clean up temp file; we're done now!
         return result;
     }
+
+    extractTextFromFile(filename: string): string {
```
demiankatz: This logic seems to be copied from the extractText method above -- so I'd suggest calling it from extractText at the appropriate point to reduce redundancy.

I also wonder if it would be better NOT to have this method truncate and delete the file -- that could be an unexpected side effect if this is used in a different context. Maybe that logic should happen outside of this method.

api/package.json
```diff
@@ -39,8 +39,8 @@
     "locutus": "^2",
     "morgan": "~1.10.1",
     "mysql2": "^3.16",
-    "n3": "^2",
-    "nanoid": "^5.1.6",
+    "n3": "^1",
```
demiankatz: Not sure why these dependencies have been downgraded here; I suspect this may be some kind of merge issue and not intended as part of the PR.

demiankatz (changes requested):
Thanks, @Jason-Benson, great work -- this looks like a solid solution to the problem. See below for the usual nitpicks and suggestions, many of which are just about possibly removing log messages that are no longer needed. :-)

api/src/services/Fedora.ts
Jason-Benson: I reworked these to include sha-512

api/src/services/TikaExtractor.ts
Jason-Benson: This whole method was left in by accident, it wasn't even called. It's been removed.

api/src/models/FedoraObject.ts
```diff
@@ -77,7 +78,29 @@ export class FedoraObject {
...
+            rs.on("data", (chunk: Buffer) => {
+                md5Hash.update(chunk);
+                sha512Hash.update(chunk);
+            });
+            rs.on("end", () => resolve());
+            rs.on("error", (err) => reject(err));
+        });
+        const md5 = md5Hash.digest("hex");
+        const sha512 = sha512Hash.digest("hex");
+        const digestHeader = `md5=${md5}, sha-512=${sha512}`;
```
demiankatz: Just for code organization purposes, it might be nice to move this block of code into a new `async computeDigestHeaderForFile(filename: string)` method and then just call `const digestHeader = this.computeDigestHeaderForFile(filename);` here.

Entirely optional, but it makes each chunk of code a little easier to mentally process!

api/src/models/FedoraObject.ts
```diff
@@ -105,7 +128,21 @@ export class FedoraObject {
             mimeType: "text/xml",
             logMessage: "Initial Ingest addDatastream - MASTER-MD",
         };
+        console.log("Getting fits MasterMetadata for file:", filename);
```
demiankatz: My earlier comment was marked as resolved, but I'd still recommend changing any `console.log` calls in this file that you want to keep to `this.log` so they write to the permanent ingest log instead of the console; that will make them more readily available for troubleshooting real-life ingests once we launch this.

demiankatz (changes requested):
Thanks for the further progress, @Jason-Benson -- looking even better now. I had one more minor idea while reviewing, and it looks like there are still a couple of comments pending from the last review, but I've gone through and resolved all the threads that are finished to clarify what remains.

api/src/services/Fedora.ts
Jason-Benson: My attempts to tinker with this very much broke ingest. How would you suggest changing this?

api/src/models/FedoraObject.ts
```diff
@@ -76,8 +77,34 @@ export class FedoraObject {
...
+                md5Hash.update(chunk);
+                sha512Hash.update(chunk);
+            });
+            rs.on("end", () => resolve());
+            rs.on("error", (err) => reject(err));
+        });
+        const md5 = md5Hash.digest("hex");
+        const sha512 = sha512Hash.digest("hex");
+        const digestHeader = `md5=${md5}, sha-512=${sha512}`;
+        return digestHeader;
```
demiankatz: We could simplify to a direct return since the variable isn't really used for anything:

```suggestion
        return `md5=${md5}, sha-512=${sha512}`;
```

api/src/services/Fedora.ts
```diff
@@ -70,7 +72,11 @@ export class Fedora {
             password: this.config.fedoraPassword,
         };
         const options = Object.assign({}, auth, _options);
-        return http(method, url, data, options);
+
+        return http(method, url, data, options).catch((err) => {
+            console.error(`Request failed for ${method.toUpperCase()} ${url}:`, err);
```
demiankatz: There are a number of `console.error` calls added to this file, and it looks like we have not console-logged errors in this layer in the past. I think it might make sense to remove them, since they seem to be associated with logic that is going to throw an exception anyway (which should lead to output elsewhere in the stack). If you think there's a strong reason to keep them around, though, I suspect they're harmless, since they should only show up in situations where things are going wrong.

demiankatz (changes requested):
See below for a couple of optional suggestions. We're definitely nearly done!

api/src/services/Fedora.ts
```diff
@@ -70,6 +72,7 @@ export class Fedora {
             password: this.config.fedoraPassword,
         };
         const options = Object.assign({}, auth, _options);
+
```
demiankatz: I don't think we need to add a blank line here (I think this was a side effect of adding some code, and then removing the added code).

```suggestion
```

demiankatz (changes requested):
Thanks, @Jason-Benson, this looks great!

See below for an incredibly trivial suggestion.

Beyond that, one other small request: can you add a line to the "Fixed" section of the [changelog](https://github.com/FalveyLibraryTechnology/VuDL/blob/dev/CHANGELOG.md) that says something like "Eliminated 2GB file limit by adding streaming support."

...and then, unless you're aware of outstanding problems, I think this can be merged!

demiankatz (approved):
Thanks, @Jason-Benson -- merging now!

PR #356, opened 2026-02-24
Next15

Jason-Benson:
Updates Next.JS to 15


demiankatz (changes requested):
Thanks, @Jason-Benson, this does seem to be working for me. A few comments, though:

1.) For some reason, the api/package-lock.json file had some unexpected things in it unrelated to the change -- maybe some local merging issue. I've reverted it to the version from the current dev branch to prevent problems.

2.) I noticed that when I started up the server, it was complaining about a missing outputFileTracingRoot configuration setting, so I just added that.

3.) One thing that concerns me slightly is that the [Next.js 15 migration notes](https://nextjs.org/docs/app/guides/upgrading/version-15) say that React 19 is now required, but we're using React 18. I'm not sure why this doesn't cause an npm conflict, or why it appears to be working correctly despite the version mismatch. Makes me a little nervous, though -- maybe it's worth seeing whether we can also bump React to 19 in this PR without breaking things, so we're running the documented compatible versions. Do you mind trying that when time permits? If it breaks stuff, maybe we should try a separate PR to upgrade React first and work through any issues, and then we can double back to this one.

demiankatz (changes requested):
Another small task on this one: I noticed while running the lint task that Next.js is complaining that the `next lint` command is deprecated in release 15 and will be removed in release 16. Apparently the project is making the move to using explicit ESLint configurations (which makes sense), and things can be automatically migrated with `npx @next/codemod@latest next-lint-to-eslint-cli .`

We'll also probably need to adjust the scripts in client/package.json, unless codemod somehow is magical enough to do that for us.

Let me know if you need any clarification!

demiankatz:
This is superseded by #358, so I'm closing it now.

PR #357, opened 2026-02-27
Depend9

Jason-Benson:
More dependency updates

### DIFF
--- a/api/src/routes/edit.ts
+++ b/api/src/routes/edit.ts
@@ -187,8 +187,8 @@ async function getChildCounts(req, res) {
     res.json(response);
 }
 
-function uploadFile(req, res, next) {
-    const { pid, stream } = req.params as { pid: string; stream: string };
+function uploadFile(req: express.Request<{ pid: string; stream: string }>, res, next) {
+    const { pid, stream } = req.params;
     const form = new IncomingForm({ multiples: true, maxFileSize: Config.getInstance().maxUploadSize });
 
     form.parse(req, async (err, fields, files) => {
@@ -213,8 +213,8 @@ edit.post(
     requireToken,
     bodyParser.json(),
     datastreamSanitizer,
-    async function (req, res) {
-        const { pid, stream } = req.params as { pid: string; stream: string };
+    async function (req: express.Request<{ pid: string; stream: string }>, res) {
+        const { pid, stream } = req.params;
         const { licenseKey } = req.body;
         try {
             const datastream = DatastreamManager.getInstance();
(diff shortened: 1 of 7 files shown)


api/src/routes/edit.ts
```diff
@@ -188,7 +188,7 @@ async function getChildCounts(req, res) {
 }
 
 function uploadFile(req, res, next) {
-    const { pid, stream } = req.params;
+    const { pid, stream } = req.params as { pid: string; stream: string };
```
demiankatz: Thanks for all the type work here -- this is great and makes things more clear.

Just one question: you mostly apply types by typing `req` in the function signature -- but in a few places like here, you instead use `as` to apply types to `req.params`. I think it's probably better to use the function signature approach everywhere, since it's a little more explicit than on-the-fly re-typing. Is there any reason you can't or shouldn't do that in cases like this one?

demiankatz (changes requested):
Thanks, @Jason-Benson, I hope to have time to look at this hands-on either later today or tomorrow. In the meantime, see below for one more theoretical question.

api/src/routes/edit.ts
Jason-Benson: That's a good idea. I did that because I was figuring it out still, but now that you point it out I definitely want to go back and change those.

api/src/routes/edit.ts
```diff
@@ -292,20 +297,25 @@ edit.post(
...
+        } catch (error) {
+            res.status(500).send(error.message);
+        }
+    },
+);
 
 edit.get("/object/:pid/datastream/:stream/metadata", requireToken, datastreamSanitizer, async (req, res) => {
     try {
-        const { pid, stream } = req.params;
+        const { pid, stream } = req.params as { pid: string; stream: string };
```
demiankatz: Looks like one "as" got missed.

api/src/routes/ingest.ts
```diff
@@ -80,29 +80,34 @@ ingest.put("/:category/:job", sanitizeParameters(), requireToken, function (req,
...
-        return res.status(404).json({ error: "Job not found" });
-    }
-    const deriv = await job.getImage(image).derivative(legalSizes[size] ?? "THUMBNAIL");
-    res.sendFile(deriv);
-});
+ingest.get(
+    "/:category/:job/:image/:size",
+    sanitizeParameters(),
+    requireToken,
+    async function (req: express.Request<{ image: string; size: string }>, res) {
```
demiankatz: Apparently it's not necessary to get the build to pass, but maybe we should include the full set of parameters anyway for completeness/accuracy:

```suggestion
    async function (req: express.Request<{ category: string; job: string; image: string; size: string }>, res) {
```

api/src/routes/ingest.ts
```diff
@@ -80,29 +80,34 @@ ingest.put("/:category/:job", sanitizeParameters(), requireToken, function (req,
...
+);
 
 ingest.delete(
     "/:category/:job/:image/\\*",
     ingest.delete(
         "/:category/:job/:image/\\*",
         sanitizeParameters({ 0: /^\*$/ }),
         requireToken,
-        async function (req, res) {
+        async function (req: express.Request<{ image: string }>, res) {
```
demiankatz: Same thought here:

```suggestion
        async function (req: express.Request<{ category: string; job: string; image: string }>, res) {
```

demiankatz (changes requested):
This is great, I'm really happy with how it is shaping up. See below for a couple final nitpicks:

demiankatz (approved):
All looking good, and working for me. I confirmed that we were already appropriately passing connection options to all BullMQ objects, so that breaking change in v5 does not impact us.

PR #358, opened 2026-03-03
Upgrade client dependencies (including Next 16/React 19).

Jason-Benson:
Working on client side dependencies


demiankatz (changes requested):
The upgraded tree view isn't working:

Image

The tree view is used to render the object type selection control when you create a new object (e.g. at `http://localhost:3000/edit/newChild`).

Sounds like there's a new unique ID requirement on the `<TreeItem>` element -- maybe this is as easy as renaming or copying the existing `nodeId` element to `id` in the code.

The upgraded date picker also isn't working:

Image

...you can test this by editing the PROCESS-MD datastream on any object. The cause of this error is less clear-cut from the message, but hopefully `TypeError: value.isValid is not a function` will point toward a particular bit of code that can be revised or updated. If you get stuck, let me know and I'll refresh my memory... or maybe looking at the component changelogs will clarify what's happening.

demiankatz (changes requested):
The lint failure in #359 led me back here to see what the lint status was in this PR... and I see that linting is failing entirely because `next lint --fix` seems to be removed in this version of Next.js.

Suggested order of operations:

1.) Maybe we should pivot to fixing types in React component definitions so that lint will pass in #359. I realize this is probably a bit of a large project, but it's also something we need to do anyway, and it may make the upgrading and testing easier in the long run. I think it's worth considering as an investment. Maybe we can start by picking one or two random components and improving their typing, to see how it works and how much effort it takes. If that goes well, we can chip away at the rest in a series of PRs. (I suggest doing only a few files per PR, so if we get into review cycles, we can work through multiple things in parallel rather than having a million conversations in a single monster PR).

2.) Maybe we should also look into replacing `next lint` with eslint configuration ahead of upgrading Next.js, so we'll have the benefit of linting while we work through the upgrade. I suspect if we fix the linting issues in this PR, we can then "backport" those changes to a separate PR that can get merged ahead of the rest of the upgrades.

What do you think?

client/eslint.config.mjs
```diff
@@ -0,0 +1,27 @@
...
+            "@typescript-eslint": typescriptEslint,
+            jsdoc,
+        },
+
+        languageOptions: {
+            parser: tsParser,
+        },
+	rules: { 
+		"@typescript-eslint/no-require-imports": "off" 
+	},
```
demiankatz: Looks like there was a mix of tabs and spaces here; converting to spaces so things line up consistently:

```suggestion
        rules: { 
            "@typescript-eslint/no-require-imports": "off" 
        },
```

client/pages/paginate/[category]/[job].tsx
```diff
@@ -2,7 +2,7 @@ import React from "react";
 import { useRouter } from "next/router";
 import JobPaginator from "../../../components/paginate/JobPaginator";
 
-export default function CategoryJob(): React.ReactElement {
+export default function CategoryJob(): React.ReactElement<any> {
```
Comment (demiankatz) [inline client/pages/paginate/[category]/[job].tsx]: I see there are still some of these `React.ReactElement<any>` types in this PR. Do we really need these? I thought we tried to separate them out to a different PR and ultimately decided to get rid of them.

client/components/edit/__snapshots__/ObjectChildCounts.test.tsx.snap
```diff
@@ -1,14 +1,7 @@
...
-    5
-     children, 
-    100
-     descendants
-  </div>,
-  ",",
-]
+<DocumentFragment>
+  {"pid":"foo:123"},
+</DocumentFragment>
```
demiankatz, line 10: I think there's a problem with this test -- it was originally rendering a `<div>` and now we're only getting the JSON part. I think for the test to be meaningful, we need to verify the message rendering.

client/components/paginate/__snapshots__/JobSelector.test.tsx.snap
```diff
@@ -1,10 +1,3 @@
...
 
-exports[`JobSelector renders 1`] = `
-[
-  "BasicBreadcrumbs",
-  <div
-    id="jobSelector"
-  />,
-]
-`;
+exports[`JobSelector renders 1`] = `<DocumentFragment />`;
```
demiankatz: Something seems wrong here -- the snapshot is empty now.

demiankatz (changes requested):
I decided to get a start at looking over this tonight -- managed to check off 57 random files out of the total of 182. I'll do more tomorrow! In any case, the majority of what I've seen so far looks great. See below for some questions and suggestions.

client/components/edit/datastream/DatastreamUploadModalContent.test.tsx
```diff
@@ -37,22 +35,22 @@ describe("DatastreamUploadModalContent", () => {
...
+        expect(asFragment()).toMatchSnapshot();
     });
 
     it("renders DatastreamLicenseContent", () => {
         editorValues.state.activeDatastream = "LICENSE";
 
-        const tree = renderer.create(<DatastreamUploadModalContent />).toJSON();
-        expect(tree).toEqual("DatastreamLicenseContent");
+        const { container } = render(<DatastreamUploadModalContent />);
+        expect(container.innerHTML).toContain("DatastreamLicenseContent");
```
demiankatz: Is there a reason that we changed `toEqual` to `toContain` here? Using `toContain` is a bit less precise; maybe it's justified if innerHTML contains messy content, but if that's the case, perhaps the messy content could be normalized with `trim()` or a regular expression to enable restoration of `toEqual`. I can live with `toContain` if we really have to, but I'd prefer to retain specificity if we can.

client/components/edit/datastream/DatastreamViewModalContent.test.tsx
```diff
@@ -54,13 +52,10 @@ describe("DatastreamViewModalContent", () => {
...
-        let tree;
-        await renderer.act(async () => {
-            tree = renderer.create(<DatastreamViewModalContent />);
-            await waitFor(() => expect(datastreamOperationValues.viewDatastream).toHaveBeenCalled());
-        });
-        expect(tree.toJSON()).toMatchSnapshot();
-        expect(mockDatatypeContent).toHaveBeenCalledWith(response);
+        const { asFragment } = render(<DatastreamViewModalContent />);
+        await waitFor(() => expect(datastreamOperationValues.viewDatastream).toHaveBeenCalled());
+        expect(asFragment()).toMatchSnapshot();
```
demiankatz: The snapshots for this test don't look right to me -- I think they're capturing the progress bar but not the actual result. Is it possible the snapshot assessment needs to be moved below the waitFor here? (That's just a guess -- maybe there's a different solution... but something doesn't seem right in the snapshot file).

demiankatz (changes requested):
See below for a couple more comments based on another partial review (I'm still just reviewing random files in an effort to chip away at the list -- have about 100 more to look at).

I also ran into a problem in hands-on testing; when I try to create a new object, it doesn't work, and I get this error:

Image

...seems to be the same issue I reported above, so that likely needs another look.

I also reviewed the date picker problem I noted above. While there is no longer a fatal error, there is a new problem: you used to be able to pick date AND time; now you can only pick date.

client/components/edit/datastream/DatastreamUploadModalContent.test.tsx
Jason-Benson: I thought there was some extra structural data in there that precluded 'toEqual', but I just reexamined it and found that we are just getting back "DatastreamAgentsContent". I switched the tests in this set to 'toEqual' and they work just fine, so I must have inherited that structure from some previous test I was working on.

Jason-Benson: Found it. In PaginatorPreview.test.tsx I had to use 'toContain' because instead of returning "preview-image" we got a bunch of extra stuff: "<div class=\"preview\"></div>"

client/components/edit/__snapshots__/StateModal.test.tsx.snap
```diff
@@ -1,642 +1,52 @@
...
-    </label>
-  </div>,
-  <button
-    onClick={[Function]}
-  >
-    Save
-  </button>,
-]
-`;
+exports[`StateModal renders correctly for a loaded object with children 1`] = `<DocumentFragment />`;
```
demiankatz: Looks like something may be wrong here -- another basically empty snapshot. Looks like it's not the only one in this file, either.

client/components/edit/children/__snapshots__/Child.test.tsx.snap
```diff
@@ -1,327 +1,11 @@
...
-          "textAlign": "right",
-        }
-      }
-    >
-      ObjectButtonBar:{"pid":"foo:123"}
-    </div>
-  </div>
-</div>
-`;
+exports[`Child handles empty titles appropriately 1`] = `<DocumentFragment />`;
```
demiankatz: Same issue here.

client/components/edit/children/__snapshots__/ChildList.test.tsx.snap
```diff
@@ -1,378 +1,74 @@
...
-  <ul
-    className="childlist__list"
-  >
-    <li>
-      Child
-    </li>
-  </ul>
-</div>
-`;
+exports[`ChildList renders using ajax-loaded root data 1`] = `<DocumentFragment />`;
```
demiankatz: ...and here.

client/components/edit/create/__snapshots__/CreateObject.test.tsx.snap
```diff
@@ -1,196 +1,170 @@
...
-      onChange={[Function]}
-      type="radio"
-      value="Deleted"
-    />
-  </label>
-  <label>
-    Select Model Type:
-    <b />
-  </label>
-  TreeItem
```
demiankatz, line 62: Any idea why the TreeItems completely disappear in this test case? I assume in other cases they are switching from three TreeItems to two TreeItems because you've restructured the test data with two top-level categories instead of three -- but I'm not sure why they're missing here.

client/components/edit/datastream/__snapshots__/DatastreamAgentsContent.test.tsx.snap
```diff
@@ -1,247 +1,5 @@
...
-    >
-      Cancel
-      <span
-        className="MuiTouchRipple-root css-8je8zh-MuiTouchRipple-root"
-      />
-    </button>
-  </div>,
-]
-`;
+exports[`DatastreamAgentsContent renders, and calls getAgents on render 1`] = `<DocumentFragment />`;
```
demiankatz: Another empty fragment.

client/components/edit/datastream/__snapshots__/DatastreamMetadataModalContent.test.tsx.snap
```diff
@@ -1,9 +1,3 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`DatastreamMetadataModalContent renders 1`] = `
-<div
-  className="MuiDialogContent-root css-lebnmz-MuiDialogContent-root"
->
-  DatatypeContent: {"data":"test1","mimeType":"test2"}
-</div>
-`;
+exports[`DatastreamMetadataModalContent renders 1`] = `<DocumentFragment />`;
```
demiankatz: Empty fragment.

client/components/edit/datastream/Datastream.test.tsx
```diff
@@ -17,12 +13,8 @@ describe("Datastream", () => {
             stream: "test1",
             disabled: true,
         };
-        // Mocking the datastream controls causes a console error; let's suppress it
-        // by mocking out console.error().
```
demiankatz, line 21: Is this test no longer generating a console error? If so, that's great -- but just want to make sure no noise has been reintroduced to the test output.

client/components/edit/datastream/DatastreamProcessMetadataContent.test.tsx
```diff
@@ -74,25 +72,31 @@ describe("DatastreamProcessMetadataContent", () => {
         render(<DatastreamProcessMetadataContent />);
 
         await waitFor(() => expect(processMetadataValues.action.setMetadata).toHaveBeenCalledWith(fakeData));
+        // if (extraStep) {
```
demiankatz: This file has a lot of commented-out code in it. Can it be cleaned up?

client/components/paginate/PaginatorPreview.test.tsx
```diff
@@ -13,14 +12,13 @@ describe("PaginatorPreview", () => {
     });
 
     it("renders", () => {
-        const tree = renderer.create(<PaginatorPreview {...props} />).toJSON();
-        expect(tree).toMatchSnapshot();
```
demiankatz, line 17: Did you intentionally remove the snapshot part of the test?

demiankatz (changes requested):
Thanks for the progress here, @Jason-Benson! I still haven't reviewed every single file, but see below for a few more questions and suggestions.

client/components/edit/datastream/Datastream.test.tsx
Jason-Benson: I think the error was somehow related to the use of react test renderer.

client/components/edit/create/__snapshots__/CreateObject.test.tsx.snap
Jason-Benson: I'm not sure why it only shows 2 treeitems per test in the new version. I just double checked the test and rearranged the logic so it runs in the same order as the original test but it made no difference.

client/components/edit/__snapshots__/StateModal.test.tsx.snap
Jason-Benson: Found the problem and corrected it. Rearranged the test to ensure that the rendering is complete before anything is done with it.

client/components/edit/children/__snapshots__/Child.test.tsx.snap
Jason-Benson: same fix

client/components/edit/children/__snapshots__/ChildList.test.tsx.snap
Jason-Benson: same fix

client/components/edit/datastream/__snapshots__/DatastreamAgentsContent.test.tsx.snap
Jason-Benson: same fix

client/components/edit/datastream/__snapshots__/DatastreamMetadataModalContent.test.tsx.snap
Jason-Benson: same fix as the others

client/components/edit/__snapshots__/ObjectOrder.test.tsx.snap
```diff
@@ -1,29 +1,5 @@
...
-   
-  <button
-    onClick={[Function]}
-  >
-    Change to 
-    custom
-  </button>
-</div>
-`;
+exports[`ObjectOrder can be aborted via confirmation dialog 1`] = `<DocumentFragment />`;
```
demiankatz: These snapshots seem wrong now -- is it possible the "act" fix broke the snapshots?

client/components/edit/children/ChildList.test.tsx
```diff
@@ -41,69 +31,97 @@ describe("ChildList", () => {
...
-            tree = getMountedChildListComponent(props);
-            await waitFor(() => expect(global.fetch).toHaveBeenCalledTimes(1));
-            expect(lastRequestUrl).toEqual("http://localhost:9000/api/edit/topLevelObjects?start=0&rows=10");
-            tree.root
-                .find((element) => {
-                    return element?.children[0] === "Show Thumbnails";
-                })
-                .props.onClick();
-        });
-        expect(tree.toJSON()).toMatchSnapshot();
```
demiankatz, line 65: Was this snapshot check intentionally removed? I noticed that the snapshot file for this test is missing some cases -- this is one of them but might not be the only one. Let me know if you need me to dig in more carefully!

client/components/edit/children/Child.test.tsx
```diff
@@ -60,58 +58,58 @@ describe("Child", () => {
...
     it("renders a thumbnail", async () => {
         props.thumbnail = true;
-        let tree;
-        await renderer.act(async () => {
-            tree = renderer.create(getChildComponent(props));
+        let asFragment;
+        await act(async () => {
+            asFragment = render(getChildComponent(props)).asFragment;
         });
+        expect(asFragment()).toMatchSnapshot();
```
demiankatz: Would it be safer to check the snapshot AFTER we've asserted that the data has been fetched, or does it not matter in this case? Just wondering about the reason for the changed order of operations here (and a few other places in this test file).

client/components/edit/create/__snapshots__/CreateObject.test.tsx.snap
demiankatz: The snapshot looks better to me now. There is a change from three tree items to two because you've changed the structure of the test data. In the old test, there were three top-level items, but you've replaced it with a more realistic hierarchy, with the items distributed between two top-level categories... so the input and output appear consistent.

client/components/edit/datastream/DatastreamProcessMetadataContent.test.tsx
```diff
@@ -133,38 +113,37 @@ describe("DatastreamProcessMetadataContent", () => {
...
+        expect(asFragment()).toMatchSnapshot();
     });
 
     it("renders a form when empty data is loaded", async () => {
-        const tree = await getRenderedTree();
+        //const tree = await getRenderedTree();
+        await renderComponent();
         expect(processMetadataValues.action.addTask).toHaveBeenCalledWith(0);
-        expect(tree).toMatchSnapshot();
+        //expect(tree).toMatchSnapshot();
```
demiankatz: Is this meant to be commented out? There's also a removed snapshot check in a later test. This is resulting in the snapshot file dropping some cases. Is that intentional?

client/components/edit/datastream/DatastreamAgentsContent.test.tsx
```diff
@@ -94,27 +92,29 @@ describe("DatastreamAgentsContent", () => {
...
-        await renderer.act(async () => {
-            tree = renderer.create(<DatastreamAgentsContent />);
+        let asFragment;
+        await act(async () => {
+            asFragment = render(<DatastreamAgentsContent />).asFragment;
         });
         await waitFor(() => expect(datastreamOperationValues.getAgents).toHaveBeenCalled());
-        await renderer.act(async () => {
-            tabChangeFunction && tabChangeFunction(null, 1);
+        await act(async () => {
```
demiankatz, line 111: Maybe it would make sense to put the `if (tabChangeFunction) {` check outside of the `await act` call -- if there's no function to call, there's no point in waiting for a callback at all. (Should work either way -- but just feels like a minor logical improvement).

There's a similar situation with pidPickerFunction in a few places below. I'll leave it to your discretion whether it's worth the effort of making the change. If you like it better as-is, feel free to just ignore me here. ;-)

client/components/edit/datastream/DatastreamMetadataModalContent.test.tsx
```diff
@@ -37,12 +36,11 @@ describe("DatastreamMetadataModalContent", () => {
             mimeType: "test2",
         };
         datastreamOperationValues.viewMetadata.mockResolvedValue(response);
-        let tree;
-        await renderer.act(async () => {
-            tree = renderer.create(<DatastreamMetadataModalContent />);
-            await waitFor(() => expect(datastreamOperationValues.viewMetadata).toHaveBeenCalled());
```
demiankatz, line 43: Are we sure we don't need this waitFor anymore? (I'll take your word for it if so -- just double-checking).

client/components/edit/ObjectStatus.test.tsx
```diff
@@ -41,23 +26,35 @@ describe("ObjectStatus", () => {
     });
 
     it("defaults to unknown state", async () => {
-        let tree;
-        await renderer.act(async () => {
-            tree = getMountedObjectStatusComponent(props);
-            await waitFor(() => expect(global.fetch).toHaveBeenCalledTimes(1));
-        });
+        const { asFragment } = render(
```
demiankatz: Is it worth refactoring this back into the original getMountedObjectStatusComponent (maybe renaming the function to getRenderedObjectStatusComponentFragment for accuracy) so we don't have to duplicate the logic in two places?

client/components/edit/ObjectSummary.test.tsx
```diff
@@ -37,11 +36,8 @@ describe("ObjectSummary", () => {
     it("displays loading message when appropriate", async () => {
         jest.spyOn(editorValues.action, "extractFirstMetadataValue").mockReturnValue("");
         const loadSpy = jest.spyOn(editorValues.action, "loadCurrentObjectDetails");
-        let tree;
-        renderer.act(() => {
-            tree = renderer.create(<ObjectSummary />);
-        });
-        expect(tree.toJSON()).toMatchSnapshot();
+        const { asFragment } = render(<ObjectSummary />);
```
demiankatz: Is it definitely correct to remove the act calls from these tests, and to check the snapshot before verifying the API call counts? (As in other places, maybe it is -- just double-checking).

client/components/edit/StateModal.test.tsx
```diff
@@ -85,47 +83,47 @@ describe("StateModal", () => {
...
         editorValues.state.objectDetailsStorage[pid] = { pid, state: "Active" };
         fetchContextValues.action.fetchJSON.mockResolvedValue({ numFound: 100 });
-        let tree;
-        await renderer.act(async () => {
-            tree = renderer.create(<StateModal />);
+        let asFragment;
+        await act(async () => {
+            asFragment = render(<StateModal />).asFragment;
         });
+        expect(asFragment()).toMatchSnapshot();
```
demiankatz: Another case of checking the snapshot before waiting for the fetch to complete (here and in another test below), which may or may not be an issue.

client/components/paginate/PaginatorPreview.test.tsx
demiankatz: This question about removing the snapshot was resolved without comment. If the answer is "yes" that's fine, but since no answer was here, I thought I'd better double-check. (This keeps coming to my attention because the associated snapshot file has been deleted as a result -- just want to be sure that's desired before I mark it as okay).

client/components/shared/DatatypeContent.test.tsx
```diff
@@ -15,36 +14,40 @@ describe("DatatypeContent", () => {
     it("renders img on image primaryType", () => {
         props.data = "test1";
         props.mimeType = "image/jpeg";
-        const tree = renderer.create(<DatatypeContent {...props} />);
 
-        expect(tree.toJSON()).toMatchSnapshot();
-        expect(tree.root.findByType("img")).toBeTruthy();
+        const { container } = render(<DatatypeContent {...props} />);
+        const { asFragment } = render(<DatatypeContent {...props} />);
```
demiankatz: Can you do this to access two properties of a single rendering, rather than creating a distinct rendering for each property? Or do they _need_ to be separated like that?

```suggestion
        const { container, asFragment } = render(<DatatypeContent {...props} />);
```

This same question applies to similar code in two additional test cases below.

demiankatz (changes requested):
Thanks, @Jason-Benson, I've done a bunch more hands-on testing and wasn't able to break anything. I also looked through all of the remaining files and spotted a handful of remaining issues (some of which may be safe to ignore, but I always err on the side of caution).

Please let me know if you want to discuss any of this stuff further -- the end is really in sight now!!

client/components/paginate/PaginatorPreview.test.tsx
Jason-Benson: I resolved that by accident. As for the snapshot test, when I originally did it I omitted the snapshot test because I thought I wasn't able to do it on a container object, but going back to it now I see I was in error and it's easy enough to add it back in. Will be in my next push.

client/components/edit/children/ChildList.test.tsx
Jason-Benson: Yeah, I see what happened. Fixing it now.

client/components/edit/children/Child.test.tsx
Jason-Benson: Today I changed the order to see if it made a difference. It didn't, but I'll leave it in that order since it more closely matches the original code before I touched it.

client/components/edit/datastream/DatastreamProcessMetadataContent.test.tsx
Jason-Benson: This took some head scratching but I think I have a good solution. I didn't meant to leave it commented out originally, I just did that while testing, meaning to go back an work out a proper solution as a later step. Well, now I guess I got to that step. I moved the snapshot test into the renderComponent function to avoid scope problems and generate a new test for each instance.

client/components/shared/DatatypeContent.test.tsx
Jason-Benson: Great idea, that works fine.

client/components/edit/StateModal.test.tsx
Jason-Benson: I changed the order of these to put the snapshot tests last. It didn't actually manifest any change in the snapshots, but I agree with the practice and will use it going forward.

client/components/edit/ObjectSummary.test.tsx
Jason-Benson: I reordered these and enclosed the renderings in await act clauses.

client/components/edit/ObjectStatus.test.tsx
Jason-Benson: Done.

client/components/edit/__snapshots__/ObjectChildCounts.test.tsx.snap
demiankatz: I think this may still be an issue -- is it possible there's some kind of timing problem here, or a missing `act`, or something? Does the new testing mechanism have trouble with contexts? I would still expect this snapshot to include text about children and descendents, so the fact that only the JSON is coming through seems to be a problem.

client/components/edit/datastream/DatastreamViewModalContent.test.tsx
demiankatz: This also still appears to be an issue.

client/components/edit/children/ChildList.test.tsx
demiankatz: Looks like snapshot checks are still missing from several tests: "allows thumbnails to be toggled on," "allows models to be toggled on," "allows child counts to be toggled on," and "renders using SelectableChild when a callback is provided" -- is that intentional?

client/components/edit/children/Child.test.tsx
demiankatz: What I'm seeing in this PR still has the order of operations changed, but your comment suggests that you realigned the code to more closely match the original order. Did something not get pushed, or is it possible that you changed one test case in the file but not all of them?

client/components/edit/datastream/DatastreamProcessMetadataContent.test.tsx
```diff
@@ -133,38 +117,35 @@ describe("DatastreamProcessMetadataContent", () => {
...
-        expect(tree).toMatchSnapshot();
+        const { asFragment } = await act(async () => render(<DatastreamProcessMetadataContent />));
+        expect(asFragment()).toMatchSnapshot();
     });
 
     it("renders a form when empty data is loaded", async () => {
-        const tree = await getRenderedTree();
+        await renderComponent();
         expect(processMetadataValues.action.addTask).toHaveBeenCalledWith(0);
-        expect(tree).toMatchSnapshot();
```
demiankatz, line 145: Is this snapshot render removed intentionally?

demiankatz: Never mind, I see now that you've centralized the snapshot rendering to the renderComponent method!

client/components/edit/ObjectStatus.test.tsx
```diff
@@ -41,23 +39,25 @@ describe("ObjectStatus", () => {
     });
 
     it("defaults to unknown state", async () => {
-        let tree;
-        await renderer.act(async () => {
-            tree = getMountedObjectStatusComponent(props);
-            await waitFor(() => expect(global.fetch).toHaveBeenCalledTimes(1));
-        });
+<<<<<<< HEAD
```
demiankatz: Looks like conflict markers ended up getting committed somehow...

client/components/edit/ObjectStatus.test.tsx
```diff
@@ -41,23 +39,25 @@ describe("ObjectStatus", () => {
...
     });
 
     it("displays the state found in the response", async () => {
         response.state = "Inactive";
-        let tree;
-        await renderer.act(async () => {
-            tree = getMountedObjectStatusComponent(props);
-            await waitFor(() => expect(global.fetch).toHaveBeenCalledTimes(1));
-        });
+<<<<<<< HEAD
```
demiankatz: Here too!

demiankatz (changes requested):
Thanks for the progress, @Jason-Benson! I've resolved some completed threads, commented on others that still have pending actions, and spotted a couple of problems highlighted below (which I think should be easy to fix -- looks like a merge conflict got committed without being fully resolved).

client/components/edit/ObjectStatus.test.tsx
Jason-Benson: D'oh!

client/components/edit/children/Child.test.tsx
Jason-Benson: I see what happened here. I reordered the rest of the tests in Child.test.tsx, but somehow forgot the very one you had flagged. Fixed.

client/components/edit/children/ChildList.test.tsx
Jason-Benson: I'm glad you pointed this out. I had meant to come back to this test since I'd learned a bunch of stuff since I initially did it. It should now match the original test much more closely.

client/components/edit/datastream/DatastreamViewModalContent.test.tsx
Jason-Benson: I reordered it and it did change the snapshot to include additional data.

client/components/edit/__snapshots__/ObjectChildCounts.test.tsx.snap
Jason-Benson: Reworking the structure of this corrected the snapshot.

client/components/edit/ObjectStatus.test.tsx
Jason-Benson: Fixed

client/components/edit/datastream/DatastreamViewModalContent.test.tsx
demiankatz: This looks better, but the "DatastreamViewModalContent renders for download-only content" snapshot is still showing a progress icon instead of the expected result. You did fix "DatastreamViewModalContent renders for viewable content" though, so we're at least halfway there!

client/components/edit/children/Child.test.tsx
demiankatz: Looking better, but I think there's just one more that needs tweaking: "renders child counts" on line 81.

client/components/edit/datastream/DatastreamAgentsContent.test.tsx
demiankatz: You haven't replied to this thread yet -- I suspect it may have gotten missed because GitHub collapsed threads and hid it.

client/components/edit/datastream/DatastreamMetadataModalContent.test.tsx
demiankatz: Same here -- I think this may still be an open question.

client/components/edit/ObjectStatus.test.tsx
```diff
@@ -41,23 +39,17 @@ describe("ObjectStatus", () => {
     });
 
     it("defaults to unknown state", async () => {
-        let tree;
-        await renderer.act(async () => {
-            tree = getMountedObjectStatusComponent(props);
-            await waitFor(() => expect(global.fetch).toHaveBeenCalledTimes(1));
-        });
+        const { asFragment } = getMountedObjectStatusComponent(props);
```
demiankatz: Do we need an await act here? The snapshot for this seems to have gone from showing an unknown status to showing the loading indicator, which makes me suspect things aren't fully loading/rendering before the snapshot is generated.

client/components/edit/datastream/DatastreamViewModalContent.test.tsx
Jason-Benson: I reordered it and I think I got better results.

client/components/edit/ObjectStatus.test.tsx
Jason-Benson: Yeah, that did it.

demiankatz (approved):
Victory at last -- thanks, @Jason-Benson, this all looks good to me now!

PR #359, opened 2026-03-11
Add <any> to React.ReactElement type definitions

Jason-Benson:
Changed all instances of ": React.ReactElement =>" to ": React.Reactelement<any> =>"


demiankatz (approved):
Thanks, @Jason-Benson!

demiankatz (changes requested):
Wait, I see that lint is failing here because of the `no-explicit-any` rule. Maybe this is not going to be so simple... We should discuss!

PR #361, opened 2026-03-31
Client typing

Jason-Benson:
Changed all instances of ": React.ReactElement<any>" to ": React.ReactElement"


demiankatz:
(Determined not to be needed after some discussion).

PR #363, opened 2026-06-16
Pdf gen iiif3

Jason-Benson:
Fixes the PDF generator which was broken by the switch to IIIF3 manifests.


demiankatz (approved):
Thanks, @Jason-Benson!

PR #365, opened 2026-06-24
Depend summer26 b

Jason-Benson:
More dependency updates


api/src/services/MetadataExtractor.ts
```diff
@@ -92,7 +92,7 @@ class MetadataExtractor {
      * @returns   Map of extracted data
      */
     public extractFedoraDetails(RDF: string): Record<string, Array<string>> {
-        const RDF_XML = this.xmlParser.parseFromString(RDF, "text/xml");
+        const RDF_XML = this.xmlParser.parseFromString(RDF, "text/xml") as unknown as Document;
```
demiankatz, line 96: Why is this necessary? The library's [internal annotations](https://github.com/xmldom/xmldom/blob/master/lib/dom-parser.js#L211) already indicate that it's returning a Document. Is there a problem with xmldom Typescript definitions not loading correctly?

client/next-env.d.ts
```diff
@@ -1,6 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
-import "./.next/dev/types/routes.d.ts";
+import "./.next/types/routes.d.ts";
```
demiankatz: This change keeps creeping into PRs. Should I commit it or revert it?

client/tsconfig.json
```diff
@@ -13,7 +13,7 @@
     "noEmit": true,
     "esModuleInterop": true,
     "module": "esnext",
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
```
demiankatz: Same question about this change: is it intentional? If not, is it good or bad?

client/next.config.js
```diff
@@ -8,4 +9,7 @@ module.exports = {
         // !! WARN !!
         ignoreBuildErrors: true,
     },
+    turbopack: {
+        root: path.join(__dirname, ".."),
```
demiankatz: From the perspective of next.js, I think the client directory should be the root, not the project root... but maybe I'm mistaken. Did you try it this way?


```suggestion
        root: __dirname,
```

demiankatz (changes requested):
Thanks, @Jason-Benson -- see below for a few questions and comments.

demiankatz (approved):
Looks good to me now!

PR #367, opened 2026-07-07
Typescript6 update

Jason-Benson:
This is the first in a set of updates to allow typescript 6. 
One notable change from TS5 is that strict is now implicit, and I updated tsconfig to match previous behavior. A lot more work is needed before ts6 can be used.

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -57,7 +57,11 @@
     "@types/express": "^5.0.6",
     "@types/express-session": "^1.18.2",
     "@types/jest": "^30",
-    "@types/node": "^25",
+    "@types/node": "^26",
+    "@types/ini": "^4.1.1",
+    "@types/jest": "^30.0.0",
+    "@types/node": "^26",
+    "@types/pdfkit": "^0.17.6",
     "@typescript-eslint/eslint-plugin": "^8.52",
     "@typescript-eslint/parser": "^8.52",
     "eslint": "^10",
@@ -69,7 +73,7 @@
     "prettier": "^3",
     "supertest": "^7",
     "ts-jest": "^29",
-    "typescript": "^5",
+    "typescript": "^6",
     "typescript-eslint": "^8.52"
   }
 }
--- a/api/package.json
+++ b/api/package.json
@@ -56,8 +56,6 @@
   "devDependencies": {
     "@types/express": "^5.0.6",
     "@types/express-session": "^1.18.2",
-    "@types/jest": "^30",
-    "@types/node": "^26",
     "@types/ini": "^4.1.1",
     "@types/jest": "^30.0.0",
     "@types/node": "^26",
--- a/api/package.json
+++ b/api/package.json
@@ -71,7 +71,7 @@
     "prettier": "^3",
     "supertest": "^7",
     "ts-jest": "^29",
-    "typescript": "^6",
+    "typescript": "^5",
     "typescript-eslint": "^8.52"
   }
 }
--- a/api/src/models/Config.ts
+++ b/api/src/models/Config.ts
@@ -37,7 +37,7 @@ class Config {
     }
 
     get backendUrl(): string {
-        return (this.ini["backend_url"] as string) ?? "http://localhost:9000";
+        return (this.ini["backend_url"] ?? "http://localhost:9000") as string;
     }
 
     get clientUrl(): string {
@@ -114,7 +114,7 @@ class Config {
     }
 
     get tikaConfigFile(): string | null {
-        return (this.ini["tika_config_file"] as string) ?? null;
+        return (this.ini["tika_config_file"] ?? null) as string | null;
     }
 
     get tikaPath(): string {
@@ -207,7 +207,7 @@ class Config {
     }
 
     get models(): Record<string, FedoraModel> {
-        return (this.ini["models"] as unknown as Record<string, FedoraModel>) || {};
+        return (this.ini["models"] as Record<string, FedoraModel>) || {};
     }
 
     get databaseSettings(): ConfigRecord {
@@ -267,7 +267,7 @@ class Config {
     }
 
     get agentDefaults(): Record<string, string> {
-        return ((this.ini["agent"] as ConfigRecord)?.["defaults"] as Record<string, string>) ?? {};
+        return ((this.ini["agent"] as ConfigRecord | undefined)?.["defaults"] as Record<string, string>) ?? {};
     }
 
     get agentRoles(): Array<string> {
--- a/api/src/models/Config.ts
+++ b/api/src/models/Config.ts
@@ -10,9 +10,9 @@ interface ConfigRecord {
 class Config {
     private static instance: Config;
 
-    protected ini;
+    protected ini: Record<string, any>;
 
-    constructor(ini: ConfigRecord) {
+    constructor(ini: Record<string, any>) {
         this.ini = ini;
     }
 
--- a/api/src/models/Config.ts
+++ b/api/src/models/Config.ts
@@ -10,9 +10,9 @@ interface ConfigRecord {
 class Config {
     private static instance: Config;
 
-    protected ini: Record<string, any>;
+    protected ini: ConfigRecord;
 
-    constructor(ini: Record<string, any>) {
+    constructor(ini: ConfigRecord) {
         this.ini = ini;
     }
 
@@ -37,117 +37,117 @@ class Config {
     }
 
     get backendUrl(): string {
-        return this.ini["backend_url"] ?? "http://localhost:9000";
+        return (this.ini["backend_url"] as string) ?? "http://localhost:9000";
     }
 
     get clientUrl(): string {
-        return this.ini["client_url"];
+        return this.ini["client_url"] as string;
     }
 
     get fedoraUsername(): string {
-        return this.ini["fedora_username"];
+        return this.ini["fedora_username"] as string;
     }
 
     get fedoraPassword(): string {
-        return this.ini["fedora_password"];
+        return this.ini["fedora_password"] as string;
     }
 
     get fedoraPidNameSpace(): string {
-        return this.ini["fedora_pid_namespace"];
+        return this.ini["fedora_pid_namespace"] as string;
     }
 
     get ffmpegPath(): string {
-        return this.ini["ffmpeg_path"];
+        return this.ini["ffmpeg_path"] as string;
     }
 
     get fitsCommand(): string {
-        return this.ini["fits_command"];
+        return this.ini["fits_command"] as string;
     }
 
     get sessionKey(): string {
-        return this.ini["session_key"] ?? "vanilla hot cocoa";
+        return (this.ini["session_key"] as string) ?? "vanilla hot cocoa";
     }
 
     get tesseractPath(): string {
-        return this.ini["tesseract_path"];
+        return this.ini["tesseract_path"] as string;
     }
 
     get tesseractAllowedChars(): string {
-        return this.ini["tesseract_allowed_characters"];
+        return this.ini["tesseract_allowed_characters"] as string;
     }
 
     get vufindUrl(): string {
-        return this.ini["vufind_url"] ?? "";
+        return (this.ini["vufind_url"] as string) ?? "";
     }
 
     get pdfDirectory(): string {
-        return this.ini["pdf_directory"];
+        return this.ini["pdf_directory"] as string;
     }
 
     get textcleanerPath(): string {
-        return this.ini["textcleaner_path"];
+        return this.ini["textcleaner_path"] as string;
     }
 
     get textcleanerSwitches(): string {
-        return this.ini["textcleaner_switches"];
+        return this.ini["textcleaner_switches"] as string;
     }
 
     get holdingArea(): string {
-        const holdingArea = this.ini["holding_area_path"];
+        const holdingArea = this.ini["holding_area_path"] as string;
         return holdingArea.endsWith("/") ? holdingArea : holdingArea + "/";
     }
 
     get ocrmypdfPath(): string {
-        return this.ini["ocrmypdf_path"];
+        return this.ini["ocrmypdf_path"] as string;
     }
 
     get processedAreaPath(): string {
-        return this.ini["processed_area_path"];
+        return this.ini["processed_area_path"] as string;
     }
 
     get restBaseUrl(): string {
-        return this.ini["base_url"];
+        return this.ini["base_url"] as string;
     }
 
     get javaPath(): string {
-        return this.ini["java_path"] ?? "java";
+        return (this.ini["java_path"] as string) ?? "java";
     }
 
-    get tikaConfigFile(): string {
-        return this.ini["tika_config_file"] ?? null;
+    get tikaConfigFile(): string | null {
+        return (this.ini["tika_config_file"] as string) ?? null;
     }
 
     get tikaPath(): string {
-        return this.ini["tika_path"];
+        return this.ini["tika_path"] as string;
     }
 
     get solrCore(): string {
-        return this.ini["solr_core"] ?? "biblio";
+        return (this.ini["solr_core"] as string) ?? "biblio";
     }
 
     get solrUrl(): string {
-        return this.ini["solr_url"] ?? "http://localhost:8983/solr";
+        return (this.ini["solr_url"] as string) ?? "http://localhost:8983/solr";
     }
 
     get solrDocumentCacheDir(): boolean | string {
-        return this.ini["solr_document_cache_dir"] ?? false;
+        return (this.ini["solr_document_cache_dir"] as string) ?? false;
     }
 
     get allowedOrigins(): string[] {
-        return this.ini["allowed_origins"] ?? [];
+        return (this.ini["allowed_origins"] as string[]) ?? [];
     }
 
     get pidNamespace(): string {
-        return this.ini["fedora_pid_namespace"] ?? "vudl";
+        return (this.ini["fedora_pid_namespace"] as string) ?? "vudl";
     }
 
     get initialPidValue(): number {
-        return parseInt(this.ini["fedora_initial_pid"] ?? "0");
+        return parseInt((this.ini["fedora_initial_pid"] as string) ?? "0");
     }
 
     get dataModels(): Record<string, string> {
         return (
-            this.ini["data_models"] ?? {
+            (this.ini["data_models"] as Record<string, string>) ?? {
                 Image: "vudl-system:ImageData",
                 PDF: "vudl-system:PDFData",
                 DOC: "vudl-system:DOCData",
@@ -161,7 +161,7 @@ class Config {
 
     get collectionModels(): Record<string, string> {
         return (
-            this.ini["collection_models"] ?? {
+            (this.ini["collection_models"] as Record<string, string>) ?? {
                 List: "vudl-system:ListCollection",
                 Resource: "vudl-system:ResourceCollection",
                 Folder: "vudl-system:FolderCollection",
@@ -170,27 +170,27 @@ class Config {
     }
 
     get institution(): string {
-        return this.ini["institution"] ?? "My University";
+        return (this.ini["institution"] as string) ?? "My University";
     }
 
     get collection(): string {
-        return this.ini["collection"] ?? "Digital Library";
+        return (this.ini["collection"] as string) ?? "Digital Library";
     }
 
     get topLevelPids(): Array<string> {
-        return this.ini["top_level_pids"] ?? [];
+        return (this.ini["top_level_pids"] as string[]) ?? [];
     }
 
     get articlesToStrip(): Array<string> {
-        return this.ini["articles_to_strip"] ?? [];
+        return (this.ini["articles_to_strip"] as string[]) ?? [];
     }
 
     get trashPid(): string | null {
-        return this.ini["trash_pid"] ?? null;
+        return (this.ini["trash_pid"] as string) ?? null;
     }
 
     get favoritePids(): Array<string> {
-        const favorites = this.ini["favorite_pids"] ?? [];
+        const favorites = (this.ini["favorite_pids"] as string[]) ?? [];
         const trash = this.trashPid;
         if (trash && !favorites.includes(trash)) {
             favorites.push(trash);
@@ -199,19 +199,19 @@ class Config {
     }
 
     get languageMap(): Record<string, string> {
-        return this.ini["LanguageMap"] ?? {};
+        return (this.ini["LanguageMap"] as Record<string, string>) ?? {};
     }
 
     get minimumValidYear(): number {
-        return parseInt(this.ini["minimum_valid_year"] ?? 1000);
+        return parseInt((this.ini["minimum_valid_year"] as string) ?? "1000");
     }
 
     get models(): Record<string, FedoraModel> {
-        return this.ini["models"] || {};
+        return (this.ini["models"] as unknown as Record<string, FedoraModel>) || {};
     }
 
     get databaseSettings(): ConfigRecord {
-        return this.ini["Database"] ?? {};
+        return (this.ini["Database"] as ConfigRecord) ?? {};
     }
 
     get databaseClient(): string {
@@ -223,7 +223,7 @@ class Config {
     }
 
     get authenticationSettings(): ConfigRecord {
-        return this.ini["Authentication"] ?? [];
+        return (this.ini["Authentication"] as ConfigRecord) ?? {};
     }
 
     get authenticationStrategy(): string {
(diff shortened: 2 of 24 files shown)


api/src/models/Config.ts
```diff
@@ -37,117 +37,117 @@ class Config {
     }
 
     get backendUrl(): string {
-        return this.ini["backend_url"] ?? "http://localhost:9000";
+        return (this.ini["backend_url"] as string) ?? "http://localhost:9000";
```
demiankatz: When there is a default fallback like this, I think we need to do the typecasting later -- otherwise we might be telling Typescript to treat `null` as a string.

I think this style would be more accurate, if it works:

```suggestion
        return (this.ini["backend_url"] ?? "http://localhost:9000") as string;
```

api/src/models/Config.ts
```diff
@@ -37,117 +37,117 @@ class Config {
...
 
     get javaPath(): string {
-        return this.ini["java_path"] ?? "java";
+        return (this.ini["java_path"] as string) ?? "java";
     }
 
-    get tikaConfigFile(): string {
-        return this.ini["tika_config_file"] ?? null;
+    get tikaConfigFile(): string | null {
+        return (this.ini["tika_config_file"] as string) ?? null;
```
demiankatz: This one is tricky since it defaults to null, but the base value may also be null. Is this legal?


```suggestion
        return (this.ini["tika_config_file"] ?? null) as string | null;
```

api/src/models/Config.ts
```diff
@@ -199,19 +199,19 @@ class Config {
...
     }
 
     get minimumValidYear(): number {
-        return parseInt(this.ini["minimum_valid_year"] ?? 1000);
+        return parseInt((this.ini["minimum_valid_year"] as string) ?? "1000");
     }
 
     get models(): Record<string, FedoraModel> {
-        return this.ini["models"] || {};
+        return (this.ini["models"] as unknown as Record<string, FedoraModel>) || {};
```
demiankatz: Why do we need the `as unknown` here but not in other places?

api/src/models/Config.ts
```diff
@@ -263,88 +263,88 @@ class Config {
     }
 
     get licenses(): Record<string, License> {
-        return this.ini["licenses"] ?? {};
+        return (this.ini["licenses"] as unknown as Record<string, License>) ?? {};
     }
 
     get agentDefaults(): Record<string, string> {
-        return this.ini?.["agent"]?.["defaults"] ?? {};
+        return ((this.ini["agent"] as ConfigRecord)?.["defaults"] as Record<string, string>) ?? {};
```
demiankatz: In cases like this, do we need to account for the possibility of an undefined value?

```suggestion
        return ((this.ini["agent"] as ConfigRecord | undefined)?.["defaults"] as Record<string, string>) ?? {};
```

api/package.json
```diff
@@ -69,7 +71,7 @@
     "prettier": "^3",
     "supertest": "^7",
     "ts-jest": "^29",
-    "typescript": "^5",
+    "typescript": "^6",
```
demiankatz: Did you mean to upgrade typescript here? I thought we were going to try to improve types using typescript 5 before actually upgrading anything.

api/src/models/ImageFile.ts
```diff
@@ -1,10 +1,7 @@
 import sharp from "sharp";
```
demiankatz, line 1: I'm not sure why these sharp changes are showing up here -- they have already been merged to dev. Maybe you need to merge upstream/dev into this branch to straighten things out.

demiankatz (changes requested):
Thanks, @Jason-Benson! I didn't have time to review this very thoroughly, but I have a few questions and comments below that might be helpful.

There is already quite a lot going on in this PR -- I'd suggest if you do any more, start separate PRs so we can discuss separate issues in separate places, and get things merged as they are finished. Otherwise, this will become hard to keep track of! (I realize it's a balancing act, because if you split things up into too many separate branches, that can lead to merge conflicts and other problems... but I think in this case there's a lot of potential for working in smaller chunks without running into that kind of trouble).

api/src/models/Config.ts
Jason-Benson: It appears that we do not.

api/src/models/Config.ts
Jason-Benson: That checks out.

api/src/models/Config.ts
Jason-Benson: That works.

api/package.json
Jason-Benson: I had forgotten about that. I backed it off to 5 and got the same test results, so I'll leave it on 5 till the currently broken tests are resolved.

api/src/models/ImageFile.ts
Jason-Benson: I tried it, but it just reported that I was already up to date.

api/src/models/Config.ts
Jason-Benson: I honestly don't know if that's needed, but it works so I'll follow your suggestion.

api/src/models/ImageFile.ts
demiankatz: Did you `git fetch upstream` before the merge? I can dig into it more deeply tomorrow if it's not that simple. :-)

api/src/models/Config.ts
```diff
@@ -37,117 +37,117 @@ class Config {
...
     }
 
     get fitsCommand(): string {
-        return this.ini["fits_command"];
+        return this.ini["fits_command"] as string;
     }
 
     get sessionKey(): string {
-        return this.ini["session_key"] ?? "vanilla hot cocoa";
+        return (this.ini["session_key"] as string) ?? "vanilla hot cocoa";
```
demiankatz: There still seem to be a few more instances of the `(var as string) ?? "default"` pattern that should be changed to `(var ?? "default") as string` to be more accurate. (Since the value might be undefined, we shouldn't typecast it before we've applied the default value, or else the type is potentially misleading).

```suggestion
        return (this.ini["session_key"] ?? "vanilla hot cocoa") as string;
```

api/src/models/Config.ts
```diff
@@ -37,117 +37,117 @@ class Config {
...
     }
 
     get solrDocumentCacheDir(): boolean | string {
-        return this.ini["solr_document_cache_dir"] ?? false;
+        return (this.ini["solr_document_cache_dir"] ?? false) as string;
     }
 
     get allowedOrigins(): string[] {
-        return this.ini["allowed_origins"] ?? [];
+        return (this.ini["allowed_origins"] as string[]) ?? [];
```
demiankatz: Same issue as before, but affecting a different type:

```suggestion
        return (this.ini["allowed_origins"] ?? []) as string[];
```

api/src/models/Config.ts
```diff
@@ -37,117 +37,117 @@ class Config {
...
 
     get initialPidValue(): number {
-        return parseInt(this.ini["fedora_initial_pid"] ?? "0");
+        return parseInt((this.ini["fedora_initial_pid"] ?? "0") as string);
     }
 
     get dataModels(): Record<string, string> {
         return (
-            this.ini["data_models"] ?? {
+            (this.ini["data_models"] as Record<string, string>) ?? {
```
demiankatz: Same issue here -- either use `Record<string, string> | undefined` at this point, or else wrap the whole null coalescing statement in front of the as clause.

api/src/models/Config.ts
```diff
@@ -161,7 +161,7 @@ class Config {
 
     get collectionModels(): Record<string, string> {
         return (
-            this.ini["collection_models"] ?? {
+            (this.ini["collection_models"] as Record<string, string>) ?? {
```
demiankatz: Same here.

api/src/models/Config.ts
```diff
@@ -170,27 +170,27 @@ class Config {
...
     }
 
     get collection(): string {
-        return this.ini["collection"] ?? "Digital Library";
+        return (this.ini["collection"] ?? "Digital Library") as string;
     }
 
     get topLevelPids(): Array<string> {
-        return this.ini["top_level_pids"] ?? [];
+        return (this.ini["top_level_pids"] as string[]) ?? [];
```
demiankatz: Here's another example of the `string[]` pattern -- there are a few more. Can you adjust them all?

api/src/models/Config.ts
```diff
@@ -199,39 +199,39 @@ class Config {
     }
 
     get languageMap(): Record<string, string> {
-        return this.ini["LanguageMap"] ?? {};
+        return (this.ini["LanguageMap"] as Record<string, string>) ?? {};
```
demiankatz: Another one of these:


```suggestion
        return (this.ini["LanguageMap"] ?? {}) as Record<string, string>;
```

api/src/models/Config.ts
```diff
@@ -199,39 +199,39 @@ class Config {
...
     }
 
     get minimumValidYear(): number {
-        return parseInt(this.ini["minimum_valid_year"] ?? 1000);
+        return parseInt((this.ini["minimum_valid_year"] ?? "1000") as string);
     }
 
     get models(): Record<string, FedoraModel> {
-        return this.ini["models"] || {};
+        return (this.ini["models"] as Record<string, FedoraModel>) || {};
```
demiankatz: Similar:


```suggestion
        return (this.ini["models"] || {}) as Record<string, FedoraModel>;
```

api/src/models/Config.ts
```diff
@@ -199,39 +199,39 @@ class Config {
...
     }
 
     get databaseSettings(): ConfigRecord {
-        return this.ini["Database"] ?? {};
+        return (this.ini["Database"] as ConfigRecord) ?? {};
```
demiankatz: Also similar:


```suggestion
        return (this.ini["Database"] ?? {}) as ConfigRecord;
```

api/src/models/Config.ts
```diff
@@ -199,39 +199,39 @@ class Config {
...
+        return (this.databaseSettings["client"] ?? "sqlite3") as string;
     }
 
     get databaseConnectionSettings(): ConfigRecord {
         return (this.databaseSettings["connection"] as ConfigRecord) ?? { filename: "./data/auth.sqlite3" };
     }
 
     get authenticationSettings(): ConfigRecord {
-        return this.ini["Authentication"] ?? [];
+        return (this.ini["Authentication"] as ConfigRecord) ?? {};
```
demiankatz: Another instance of this pattern; please fix 'em all! :-)

api/src/models/Config.ts
```diff
@@ -242,109 +242,109 @@ class Config {
...
     }
 
     get samlEntryPoint(): string {
-        return (this.authenticationSettings["saml_entry_point"] as string) ?? "";
+        return (this.authenticationSettings["saml_entry_point"] ?? "") as string;
     }
 
     get licenses(): Record<string, License> {
-        return this.ini["licenses"] ?? {};
+        return (this.ini["licenses"] as unknown as Record<string, License>) ?? {};
```
demiankatz: A different variation:

```suggestion
        return (this.ini["licenses"] ?? {}) as Record<string, License>;
```

api/src/models/Config.ts
```diff
@@ -242,109 +242,109 @@ class Config {
...
     }
 
     get agentDefaults(): Record<string, string> {
-        return this.ini?.["agent"]?.["defaults"] ?? {};
+        return ((this.ini["agent"] as ConfigRecord | undefined)?.["defaults"] as Record<string, string>) ?? {};
```
demiankatz: Another more complex scenario:

```suggestion
        return ((this.ini["agent"] as ConfigRecord | undefined)?.["defaults"] ?? {}) as Record<string, string>;
```

api/src/models/Config.ts
```diff
@@ -242,109 +242,109 @@ class Config {
...
     }
 
     get agentRoles(): Array<string> {
-        return this.ini?.["agent"]?.["roles"] ?? [];
+        return ((this.ini["agent"] as ConfigRecord)?.["roles"] as string[]) ?? [];
```
demiankatz: This needs to match the pattern of agentDefaults:


```suggestion
        return ((this.ini["agent"] as ConfigRecord | undefined)?.["roles"] ?? []) as string[];
```

api/src/models/Config.ts
```diff
@@ -242,109 +242,109 @@ class Config {
...
     }
 
     get agentTypes(): Array<string> {
-        return this.ini?.["agent"]?.["types"] ?? [];
+        return ((this.ini["agent"] as ConfigRecord)?.["types"] as string[]) ?? [];
```
demiankatz: This needs to match the above case. There are also several more `... as ConfigRecord)?.` cases below that need similar reformatting.

api/src/models/Config.ts
```diff
@@ -242,109 +242,109 @@ class Config {
...
     }
 
     get dublinCoreFields(): Record<string, Record<string, string | Array<string>>> {
-        return this.ini?.["dublin_core"] ?? {};
+        return (this.ini["dublin_core"] as Record<string, Record<string, string | Array<string>>>) ?? {};
```
demiankatz: Another oddball case:

```suggestion
        return (this.ini["dublin_core"] ?? {}) as Record<string, Record<string, string | Array<string>>>;
```

api/src/models/Config.ts
```diff
@@ -242,109 +242,109 @@ class Config {
...
     }
 
     get redisLockDuration(): number {
-        return parseInt(this.ini?.["queue"]?.["lockDuration"] ?? "30000");
+        return parseInt(((this.ini["queue"] as ConfigRecord)?.["lockDuration"] ?? "30000") as string);
     }
 
     get processMetadataDefaults(): Record<string, string> {
-        return this.ini?.["process_metadata_defaults"] ?? {};
+        return (this.ini["process_metadata_defaults"] as Record<string, string>) ?? {};
```
demiankatz: And another variation:

```suggestion
        return (this.ini["process_metadata_defaults"] ?? {}) as Record<string, string>;
```

api/src/models/Config.ts
```diff
@@ -242,109 +242,109 @@ class Config {
...
     }
 
     get toolPresets(): Array<Record<string, string>> {
-        return this.ini?.["tool_presets"] ?? [];
+        return (this.ini["tool_presets"] as unknown as Array<Record<string, string>>) ?? [];
```
demiankatz: And another one that can be simplified:


```suggestion
        return (this.ini["tool_presets"] ?? []) as Array<Record<string, string>>;
```

api/src/models/Config.ts
```diff
@@ -242,109 +242,109 @@ class Config {
...
-        const pixelLimit = this.ini?.["sharp"]?.["limitInputPixels"] ?? "268402689";
+        const pixelLimit = ((this.ini["sharp"] as ConfigRecord)?.["limitInputPixels"] ?? "268402689") as string;
         return {
             limitInputPixels: parseInt(pixelLimit),
         };
     }
 
     get max409Retries(): number {
-        return this.ini["max_409_retries"] ?? 3;
+        return (this.ini["max_409_retries"] as unknown as number) ?? 3;
```
demiankatz: This should work a little more tidily:

```suggestion
        return (this.ini["max_409_retries"] ?? 3) as number;
```

.... though if the ini parser is actually returning strings, maybe we should introduce the parseInt pattern here like we use in some other places.

api/src/models/Job.ts
```diff
@@ -1,5 +1,20 @@
 import { createWriteStream, openSync, closeSync, existsSync as fileExists, statSync } from "fs";
 import PDFDocument = require("pdfkit");
+
+declare global {
```
demiankatz: Where/how is this used? Seems kind of hacky, and I don't actually see it being referenced anywhere. Am I missing something? Do we need to contribute upstream improvements to `@types/pdfkit`?

api/src/models/Config.ts
Jason-Benson: When I tried this out it flagged it as problematic with the following message:
Conversion of type 'ConfigValue' to type 'Record<string, string>[]' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Type 'string[]' is not comparable to type 'Record<string, string>[]'.
    Type 'string' is not comparable to type 'Record<string, string>'.ts(2352)

demiankatz: I think the problem may actually be that our ConfigValue type is not flexible enough. See:

https://github.com/FalveyLibraryTechnology/VuDL/blob/6b204fbd7cdd338be643c917d70cfd46635183fb/api/src/models/Config.ts#L5

I think we may need to change that to:

```
type ConfigValue = string | string[] | ConfigRecord | ConfigRecord[];
```

As defined, the type doesn't allow an array of records, but the tool_presets in the existing configuration does in fact return an array of records -- so I think the problem is with our type! If you tweak that, then the typecasting may actually work as expected.

api/src/models/Job.ts
Jason-Benson: This was a work around for a flaw in @type/pdfkit which was revealed with the TS5 update. pdfObj.openImage was a valid method, but it wasn't declared in @type/pdfkit, so it was getting flagged. That being said, I have found another (possibly less dumb) way to correct for this.

api/src/types/pdfkit-augment.d.ts
```diff
@@ -0,0 +1,15 @@
+// Work around for a flaw in @types/pdfkit, which omits `openImage()` from the type definitions.
```
demiankatz, line 1: Thanks, this is definitely an easier-to-understand version of the fix, and a reasonable solution while we wait for an upstream fix.

If we're impatient for an upstream fix, I wonder if we should try submitting a PR against https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/pdfkit/index.d.ts -- it appears that this repository is fairly active and people have been pushing changes to the pdfkit definitions relatively recently, so more fixes may be accepted.

api/src/models/Config.ts
```diff
@@ -37,117 +37,117 @@ class Config {
...
     }
 
     get solrUrl(): string {
-        return this.ini["solr_url"] ?? "http://localhost:8983/solr";
+        return (this.ini["solr_url"] ?? "http://localhost:8983/solr") as string;
     }
 
     get solrDocumentCacheDir(): boolean | string {
-        return this.ini["solr_document_cache_dir"] ?? false;
+        return (this.ini["solr_document_cache_dir"] ?? false) as string;
```
demiankatz: Looks like we need to account for the `false` value here:

```suggestion
        return (this.ini["solr_document_cache_dir"] ?? false) as boolean | string;
```

api/src/models/Config.ts
```diff
@@ -37,117 +37,117 @@ class Config {
...
 
     get initialPidValue(): number {
-        return parseInt(this.ini["fedora_initial_pid"] ?? "0");
+        return parseInt((this.ini["fedora_initial_pid"] ?? "0") as string);
     }
 
     get dataModels(): Record<string, string> {
         return (
-            this.ini["data_models"] ?? {
+            (this.ini["data_models"] as Record<string, string>) ?? {
```
demiankatz: Since we're using `??`, this value might be undefined, so I think we need to reflect that in the type to be entirely accurate.

```suggestion
            (this.ini["data_models"] as Record<string, string> | undefined) ?? {
```

api/src/models/Config.ts
demiankatz: Same here:

```suggestion
            (this.ini["collection_models"] as Record<string, string> | undefined) ?? {
```

api/src/models/Config.ts
demiankatz: This pattern still needs to be fixed in several places below:

```suggestion
        return (this.ini["top_level_pids"] ?? [])  as string[];
```

api/src/models/Config.ts
```diff
@@ -170,27 +170,27 @@ class Config {
...
     }
 
     get topLevelPids(): Array<string> {
-        return this.ini["top_level_pids"] ?? [];
+        return (this.ini["top_level_pids"] as string[]) ?? [];
     }
 
     get articlesToStrip(): Array<string> {
-        return this.ini["articles_to_strip"] ?? [];
+        return (this.ini["articles_to_strip"] as string[]) ?? [];
```
demiankatz:
```suggestion
        return (this.ini["articles_to_strip"] ?? []) as string[];
```

api/src/models/Config.ts
```diff
@@ -170,27 +170,27 @@ class Config {
...
     }
 
     get trashPid(): string | null {
-        return this.ini["trash_pid"] ?? null;
+        return (this.ini["trash_pid"] ?? null) as string;
```
demiankatz: This can still be null:

```suggestion
        return (this.ini["trash_pid"] ?? null) as string | null;
```

api/src/models/Config.ts
```diff
@@ -170,27 +170,27 @@ class Config {
...
     }
 
     get favoritePids(): Array<string> {
-        const favorites = this.ini["favorite_pids"] ?? [];
+        const favorites = (this.ini["favorite_pids"] as string[]) ?? [];
```
demiankatz:
```suggestion
        const favorites = (this.ini["favorite_pids"] ?? []) as string[];
```

api/src/models/Config.ts
demiankatz:
```suggestion
        return (this.ini["LanguageMap"] ?? {}) as Record<string, string>;
```

api/src/models/Config.ts
demiankatz:
```suggestion
        return (this.ini["models"] || {}) as Record<string, FedoraModel>;
```

api/src/models/Config.ts
demiankatz:
```suggestion
        return (this.ini["Database"] ?? {}) as ConfigRecord;
```

api/src/models/Config.ts
demiankatz: Also note that `databaseConnectionSettings` above this function is typecasting too early and should be refactored similarly. That's an existing problem, but we might as well fix it here!

```suggestion
        return (this.ini["Authentication"] ?? {}) as ConfigRecord;
```

api/src/models/Config.ts
demiankatz: I don't see an obvious reason why we need the extra `as unknown` here, but let me know if I'm missing something!

```suggestion
        return (this.ini["licenses"] ?? {}) as Record<string, License>;
```

api/src/models/Config.ts
demiankatz:
```suggestion
        return ((this.ini["agent"] as ConfigRecord | undefined)?.["defaults"] ?? {}) as Record<string, string>;
```

api/src/models/Config.ts
demiankatz:
```suggestion
        return ((this.ini["agent"] as ConfigRecord)?.["roles"] ?? []) as string[];
```

api/src/models/Config.ts
demiankatz:
```suggestion
        return ((this.ini["agent"] as ConfigRecord)?.["types"] ?? []) as string[];
```

api/src/models/Config.ts
```diff
@@ -242,109 +242,109 @@ class Config {
...
     }
 
     get dublinCoreFields(): Record<string, Record<string, string | Array<string>>> {
-        return this.ini?.["dublin_core"] ?? {};
+        return (this.ini["dublin_core"] ?? {}) as Record<string, Record<string, string | Array<string>>>;
     }
 
     get redisConnectionSettings(): Record<string, string> {
-        return this.ini?.["queue"]?.["connection"] ?? {};
+        return ((this.ini["queue"] as ConfigRecord)?.["connection"] as Record<string, string>) ?? {};
```
demiankatz:
```suggestion
        return ((this.ini["queue"] as ConfigRecord | undefined)?.["connection"] as ?? {}) Record<string, string>;
```

api/src/models/Config.ts
```diff
@@ -242,109 +242,109 @@ class Config {
...
     }
 
     get redisDefaultQueueName(): string {
-        return this.ini?.["queue"]?.["defaultQueueName"] ?? "vudl";
+        return ((this.ini["queue"] as ConfigRecord)?.["defaultQueueName"] ?? "vudl") as string;
```
demiankatz:
```suggestion
        return ((this.ini["queue"] as ConfigRecord | undefined)?.["defaultQueueName"] ?? "vudl") as string;
```

api/src/models/Config.ts
```diff
@@ -242,109 +242,109 @@ class Config {
...
     }
 
     get redisQueueJobMap(): Record<string, string> {
-        return this.ini?.["queue"]?.["jobMap"] ?? {};
+        return ((this.ini["queue"] as ConfigRecord)?.["jobMap"] as Record<string, string>) ?? {};
```
demiankatz:
```suggestion
        return ((this.ini["queue"] as ConfigRecord | undefined)?.["jobMap"] ?? {}) as Record<string, string>;
```

api/src/models/Config.ts
```diff
@@ -242,109 +242,109 @@ class Config {
...
     }
 
     get redisLockDuration(): number {
-        return parseInt(this.ini?.["queue"]?.["lockDuration"] ?? "30000");
+        return parseInt(((this.ini["queue"] as ConfigRecord)?.["lockDuration"] ?? "30000") as string);
```
demiankatz:
```suggestion
        return parseInt(((this.ini["queue"] as ConfigRecord | undefined)?.["lockDuration"] ?? "30000") as string);
```

api/src/models/Config.ts
```diff
@@ -242,109 +242,109 @@ class Config {
...
     }
 
     get toolPresets(): Array<Record<string, string>> {
-        return this.ini?.["tool_presets"] ?? [];
+        return (this.ini["tool_presets"] ?? []) as Array<Record<string, string>>;
     }
 
     get sharpOptions(): Record<string, unknown> {
-        const pixelLimit = this.ini?.["sharp"]?.["limitInputPixels"] ?? "268402689";
+        const pixelLimit = ((this.ini["sharp"] as ConfigRecord)?.["limitInputPixels"] ?? "268402689") as string;
```
demiankatz:
```suggestion
        const pixelLimit = ((this.ini["sharp"] as ConfigRecord | undefined)?.["limitInputPixels"] ?? "268402689") as string;
```

api/src/models/Config.ts
```diff
@@ -242,109 +242,109 @@ class Config {
...
     }
 
     get max409Retries(): number {
-        return this.ini["max_409_retries"] ?? 3;
+        return (this.ini["max_409_retries"] ?? 3) as number;
     }
 
     get maxUploadSize(): number {
-        return this.ini?.["upload"]?.["sizeLimit"] ?? 200 * 1024 * 1024;
+        return ((this.ini["upload"] as ConfigRecord)?.["sizeLimit"] as unknown as number) ?? 200 * 1024 * 1024;
```
demiankatz: This probably makes more sense if we parse an integer from a string. I moved the math into a comment and used a hard-coded number to simplify:

```suggestion
        // Default to 200MB (1024 * 1024 * 200 = 209715200)
        return parseInt(((this.ini["upload"] as ConfigRecord | undefined)?.["sizeLimit"] ?? "209715200") as string);
```

api/src/models/Config.ts
```diff
@@ -242,109 +242,109 @@ class Config {
...
     }
 
     get notifyMethod(): string {
-        return this.ini?.["notify"]?.["method"] ?? "ntfy";
+        return ((this.ini["notify"] as ConfigRecord)?.["method"] ?? "ntfy") as string;
```
demiankatz:
```suggestion
        return ((this.ini["notify"] as ConfigRecord | undefined)?.["method"] ?? "ntfy") as string;
```

api/src/models/Config.ts
```diff
@@ -242,109 +242,109 @@ class Config {
...
     }
 
     get ntfyConfig(): Record<string, string> {
         return {
-            defaultChannel: this.ini?.["notify"]?.["ntfy_defaultChannel"] ?? "vudl-ntfy",
+            defaultChannel: ((this.ini["notify"] as ConfigRecord)?.["ntfy_defaultChannel"] ?? "vudl-ntfy") as string,
```
demiankatz:
```suggestion
            defaultChannel: ((this.ini["notify"] as ConfigRecord | undefined)?.["ntfy_defaultChannel"] ?? "vudl-ntfy") as string,
```

api/src/models/Config.ts
```diff
@@ -242,109 +242,109 @@ class Config {
...
         };
     }
 
     get indexerLockRetries(): number {
-        return parseInt(this.ini?.["indexer"]?.["lockRetries"] ?? 60);
+        return parseInt(((this.ini["indexer"] as ConfigRecord)?.["lockRetries"] ?? "60") as string);
```
demiankatz:
```suggestion
        return parseInt(((this.ini["indexer"] as ConfigRecord | undefined)?.["lockRetries"] ?? "60") as string);
```

api/src/models/Config.ts
```diff
@@ -242,109 +242,109 @@ class Config {
...
     }
 
     get indexerLockWaitMs(): number {
-        return parseInt(this.ini?.["indexer"]?.["lockWaitMs"] ?? 1000);
+        return parseInt(((this.ini["indexer"] as ConfigRecord)?.["lockWaitMs"] ?? "1000") as string);
```
demiankatz:
```suggestion
        return parseInt(((this.ini["indexer"] as ConfigRecord | undefined)?.["lockWaitMs"] ?? "1000") as string);
```

api/src/models/Config.ts
```diff
@@ -242,109 +242,109 @@ class Config {
...
     }
 
     get indexerExceptionRetries(): number {
-        return parseInt(this.ini?.["indexer"]?.["exceptionRetries"] ?? 10);
+        return parseInt(((this.ini["indexer"] as ConfigRecord)?.["exceptionRetries"] ?? "10") as string);
```
demiankatz:
```suggestion
        return parseInt(((this.ini["indexer"] as ConfigRecord | undefined)?.["exceptionRetries"] ?? "10") as string);
```

api/src/models/Config.ts
```diff
@@ -242,109 +242,109 @@ class Config {
...
     }
 
     get indexerExceptionWaitMs(): number {
-        return parseInt(this.ini?.["indexer"]?.["exceptionWaitMs"] ?? 500);
+        return parseInt(((this.ini["indexer"] as ConfigRecord)?.["exceptionWaitMs"] ?? "500") as string);
```
demiankatz:
```suggestion
        return parseInt(((this.ini["indexer"] as ConfigRecord | undefined)?.["exceptionWaitMs"] ?? "500") as string);
```

api/src/models/Config.ts
```diff
@@ -242,109 +242,112 @@ class Config {
...
     }
 
     get dublinCoreFields(): Record<string, Record<string, string | Array<string>>> {
-        return this.ini?.["dublin_core"] ?? {};
+        return (this.ini["dublin_core"] ?? {}) as Record<string, Record<string, string | Array<string>>>;
     }
 
     get redisConnectionSettings(): Record<string, string> {
-        return this.ini?.["queue"]?.["connection"] ?? {};
+        return ((this.ini["queue"] as ConfigRecord)?.["connection"] as Record<string, string>) ?? {};
```
demiankatz: Sorry, I see that my previous suggestion here was garbled. Let me try again:

```suggestion
        return ((this.ini["queue"] as ConfigRecord | undefined)?.["connection"]  ?? {}) as Record<string, string>;
```

api/src/models/Config.ts
```diff
@@ -242,109 +242,112 @@ class Config {
...
     }
 
     get samlEntryPoint(): string {
-        return (this.authenticationSettings["saml_entry_point"] as string) ?? "";
+        return (this.authenticationSettings["saml_entry_point"] ?? "") as string;
     }
 
     get licenses(): Record<string, License> {
-        return this.ini["licenses"] ?? {};
+        return (this.ini["licenses"] as unknown as Record<string, License>) ?? {};
```
demiankatz: This one still needs a little work. If my previous suggestions were unsuccessful, maybe we have to settle for this one. It could be that Typescript can't figure out how to resolve a ConfigValue into a License -- but even if we can't avoid `as unknown`, I think we should do our null coalescing before our type juggling to keep the order of operations more true to life.

```suggestion
        return (this.ini["licenses"]  ?? {}) as unknown as Record<string, License>;
```

api/src/models/Config.ts
```diff
@@ -242,109 +242,112 @@ class Config {
...
     }
 
     get agentDefaults(): Record<string, string> {
-        return this.ini?.["agent"]?.["defaults"] ?? {};
+        return ((this.ini["agent"] as ConfigRecord | undefined)?.["defaults"] ?? {}) as Record<string, string>;
     }
 
     get agentRoles(): Array<string> {
-        return this.ini?.["agent"]?.["roles"] ?? [];
+        return ((this.ini["agent"] as ConfigRecord)?.["roles"] ?? []) as string[];
```
demiankatz: Might be undefined:

```suggestion
        return ((this.ini["agent"] as ConfigRecord | undefined)?.["roles"] ?? []) as string[];
```

api/src/models/Config.ts
```diff
@@ -242,109 +242,112 @@ class Config {
...
     }
 
     get agentTypes(): Array<string> {
-        return this.ini?.["agent"]?.["types"] ?? [];
+        return ((this.ini["agent"] as ConfigRecord)?.["types"] ?? []) as string[];
```
demiankatz:
```suggestion
        return ((this.ini["agent"] as ConfigRecord | undefined)?.["types"] ?? []) as string[];
```

api/src/models/Config.ts
```diff
@@ -242,109 +242,112 @@ class Config {
...
+        const pixelLimit = ((this.ini["sharp"] as ConfigRecord | undefined)?.["limitInputPixels"] ??
+            "268402689") as string;
         return {
             limitInputPixels: parseInt(pixelLimit),
         };
     }
 
     get max409Retries(): number {
-        return this.ini["max_409_retries"] ?? 3;
+        return (this.ini["max_409_retries"] ?? 3) as number;
```
demiankatz: I think this is probably more accurate:

```suggestion
        return parseInt(this.ini["max_409_retries"] ?? "3" as string);
```

Jason-Benson: Needed a slight tweak, an extra set of parenthesis:
return parseInt((this.ini["max_409_retries"] ?? "3") as string);

demiankatz: Good catch!

api/src/models/Config.ts
```diff
@@ -242,109 +242,113 @@ class Config {
...
+        const pixelLimit = ((this.ini["sharp"] as ConfigRecord | undefined)?.["limitInputPixels"] ??
+            "268402689") as string;
         return {
             limitInputPixels: parseInt(pixelLimit),
         };
     }
 
     get max409Retries(): number {
-        return this.ini["max_409_retries"] ?? 3;
+        // return (this.ini["max_409_retries"] ?? 3) as number;
```
demiankatz: I don't think we need this commented-out line! Feel free to just accept my suggestion here to get rid of it without having to go through the process in your local environment; I am confident it won't have any side effects.

```suggestion
```

I'll do some hands-on testing of this next week, and then I'm pretty sure we'll be able to merge it! Thanks again!

